Issue #2, opened 2016-11-15
Decide on a naming scheme for "important" iterations on the OpenTracing specification


bensigelman:
This has come up a few times, most recently [here](https://github.com/opentracing/opentracing.io/issues/161). Using semver for the OpenTracing spec itself would be excessively confusing... there would be (a) the semver for the OT spec, (b) the semver for the particular OT language API, and (c) the semver for the tracing implementation. Too many numbers!

I would rather come up with an ordered naming scheme for "important" (*) OpenTracing specification iterations... maybe alphabetical nouns in a given category. (C.f. Ubuntu's naming scheme or whatever)

(*): Not sure what "important" means... doesn't just have to be "breaking" changes to the spec.

wu-sheng:
@yurishkuro @cwe1ss 
continue to discuss.

As @cwe1ss said,  using the same version maybe not a good idea. But as a spec, OT can use a single number( not include dot in version number), like OT v1, OT v2, etc. OT spec won't have a bug or a reflection needs

Fot the particular OT  language API, they use their self define version num, or traditional version num, only to use the first version num as same as OT spec. such as: OpenTracing-java impl can use 1.x.x as a version number, which declares to support OT v1 spec.

This is very easy to understand for all users. This type of version naming is something like JDK spec. 

And for `(c) the semver for the tracing implementation`, it can wait, let every team to decide. Any team's homepage will show the relation between  `tracing implementation`'s version and OT version.

bensigelman:
We could also do roman numerals for the OT spec version... `OpenTracing I`, `OpenTracing II`, etc?

wu-sheng:
Roman num is very clear when it's less than 12. like `Ⅰ Ⅱ  Ⅲ ... Ⅹ ... Ⅻ`. May be it's good enough for now?

Too large num may be not so pretty?

bensigelman:
I think the "Roman Numerals" idea is the first one I've actually felt excited about for more than 24 hours. @yurishkuro what do you think?

yurishkuro:
I suppose I am not so worried about the exact naming/numbering as about what we intend to use it for. For example, how do the following changes affect the versioning?
  * correcting a typo
  * rephrasing a sentence
  * fixing incorrect sentence
  * adding new standard tag/field
  * changing semantic restrictions (e.g. something about baggage encoding or propagation)
  * changing conceptual API (e.g. the change when we moved get/set baggage from Context to Span)

[...]

PR #20, opened 2016-11-28
Record version 1.0 of the OpenTracing specification

bensigelman:
This is a slightly more formal and less language-specific "spelling" of the content at opentracing.io/spec.

Once this is merged, the opentracing.io/spec content will still exist but will mostly be a link to this document and some short explanatory / contextual text with links to other areas of the docs site.

I also want to create .yaml versions of the "official" standard tags and (hopefully) standard structured log keys, include them in this repo, and link to them from this spec.

cc @yurishkuro @wu-sheng (who both commented on the branch pre-PR), as well as @adriancole @jmacd @cwe1ss @dawallin @codingfabian @basvanbeek @beberlei @dkuebric @lookfwd @michaelsembwever who might want to weigh in.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -52,12 +52,15 @@ Temporal relationships between Spans in a single Trace
 
 Each **Span** encapsulates the following state:
 
+- An operation name
 - A start timestamp
 - A finish timestamp
 - A set of zero or more key:value **Span Tags**. The keys must be strings. The
   values may be strings, bools, or numeric types.
 - A set of zero or more **Span Logs**, each of which is itself a key:value map
-  paired with a timestamp
+  paired with a timestamp. The keys must be strings, though the values may be
+  of any type. Not all OpenTracing implementations must support every value
+  type.
 - A **SpanContext** (see below)
 - **References** to zero or more causally-related **Spans** (via the
   **SpanContext** of those related **Spans**)
@@ -88,6 +91,14 @@ Required parameters
       larger computation). The operation name should be **the most general
       string that identifies a (statistically) interesting class of `Span`
       instances**. That is, `"get_user"` is better than `"get_user/314159"`.
+
+For example, here are potential **operation names** for a `Span` that gets hypothetical account information:
+
+| Operation Name | Guidance |
+|:---------------|:--------|
+| `get` | Too general |
+| `get_account/792` | Too specific |
+| `get_account` | Good, and `account_id=792` would make a nice **`Span` tag** |
 
 Optional parameters
     - Zero or more **references**, including a shorthand for `CHILD_OF` and
@@ -128,14 +139,21 @@ With the exception of the method to retrieve the `Span`'s `SpanContext`, none of
 
 There should be no parameters.
 
-**Returns** the `SpanContext` for the given `Span`
+**Returns** the `SpanContext` for the given `Span`. The returned value may be used even after the `Span` is finished.
 
 #### Overwrite the operation name
 
 Required parameters
     - The new **operation name**, which supersedes whatever was passed in when the `Span` was started
 
-#### Set `Span` tags
+#### Finish the `Span`
+
+Optional parameters
+    - An explicit **finish timestamp** for the `Span`; if omitted, the current walltime is used implicitly.
+
+With the exception of the method to retrieve a `Span`'s `SpanContext`, no method may be called on a `Span` instance after it's finished.
+
+#### Set a `Span` tag
 
 Required parameters
     - The tag key, which must be a string
--- a/specification.md
+++ b/specification.md
@@ -79,7 +79,7 @@ The `Tracer` interface creates `Span`s and understands how to `Inject`
 (serialize) and `Extract` (deserialize) them across process boundaries.
 Formally, it has the following capabilities:
 
-#### Starting a new `Span`
+#### Start a new `Span`
 
 Required parameters
     - An **operation name**, a human-readable string which
@@ -98,14 +98,14 @@ Optional parameters
 
 **Returns** a `Span` instance that's already started (but not `Finish`ed)
 
-#### Injecting a `SpanContext` into a carrier
+#### Inject a `SpanContext` into a carrier
 
 Required parameters
     - A **`SpanContext`** instance
     - A **format** descriptor (typically but not necessarily a string constant) which tells the `Tracer` implementation how to encode the `SpanContext` in the carrier parameter
     - A **carrier**, whose type is dictated by the **format**. The `Tracer` implementation will encode the `SpanContext` in this carrier object according to the **format**.
 
-#### Extracting a `SpanContext` from a carrier
+#### Extract a `SpanContext` from a carrier
 
 Required parameters
     - A **format** descriptor (typically but not necessarily a string constant) which tells the `Tracer` implementation how to decode `SpanContext` from the carrier parameter
@@ -124,18 +124,18 @@ Both injection and extraction rely on an extensible **format** parameter that di
 
 With the exception of the method to retrieve the `Span`'s `SpanContext`, none of the below may be called after the `Span` is finished.
 
-#### Retrieving the `Span`s `SpanContext`
+#### Retrieve the `Span`s `SpanContext`
 
 There should be no parameters.
 
 **Returns** the `SpanContext` for the given `Span`
 
-#### Overwriting the operation name
+#### Overwrite the operation name
 
 Required parameters
     - The new **operation name**, which supersedes whatever was passed in when the `Span` was started
 
-#### Setting `Span` tags
+#### Set `Span` tags
 
 Required parameters
     - The tag key, which must be a string
--- a/specification.md
+++ b/specification.md
@@ -1,0 +1,191 @@
+# The OpenTracing Semantic Specification
+
+**Version:** 1.0
+
+## Document Overview
+
+This is the "formal" OpenTracing semantic specification. Since OpenTracing must work across many languages, this document takes care to avoid language-specific concepts. That said, there is an understanding throughout that all languages have some concept of an "interface" which encapsulates a set of related capabilities.
+
+## The OpenTracing Data Model
+
+**Traces** in OpenTracing are defined implicitly by their **Spans**. In
+particular, a **Trace** can be thought of as a directed acyclic graph (DAG) of
+**Spans**, where the edges between **Spans** are called **References**.
+
+For example, the following is an example **Trace** made up of 8 **Spans**:
+
+~~~
+Causal relationships between Spans in a single Trace
+
+
+        [Span A]  ←←←(the root span)
+            |
+     +------+------+
+     |             |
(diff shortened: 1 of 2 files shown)


specification.md
```diff
@@ -0,0 +1,191 @@
...
+
+### `Span`
+
+With the exception of the method to retrieve the `Span`'s `SpanContext`, none of the below may be called after the `Span` is finished.
+
+#### Retrieve the `Span`s `SpanContext`
+
+There should be no parameters.
+
+**Returns** the `SpanContext` for the given `Span`
```
yurishkuro: Maybe add: The returned value can be used as span reference even after the original span is Finished.

specification.md
```diff
@@ -0,0 +1,191 @@
...
+
+#### Overwrite the operation name
+
+Required parameters
+    - The new **operation name**, which supersedes whatever was passed in when the `Span` was started
+
+#### Set `Span` tags
```
yurishkuro: s/tags/tag/

specification.md
```diff
@@ -0,0 +1,191 @@
...
+    - One or more key:value pairs, where the keys must be strings and the values may have any type at all. Some OpenTracing implementations may handle more (or more of) certain log values than others.
+
+Optional parameters
+    - An explicit timestamp. If specified, it must fall between the local start and finish time for the span.
+
+Note that the OpenTracing project documents certain **"standard log keys"** which have prescribed semantic meanings. TODO: add these log keys to a central .yaml file and link here.
+
+#### Set a **baggage** item
+
+Baggage items are key:value string pairs that apply to the given `Span`, its `SpanContext`, and **all `Spans` which directly or transitively _reference_ the local `Span`.** That is, baggage items propagate in-band along with the trace itself.
```
yurishkuro: should this clarify that it only applies to childOf/followsFrom references? Granted, there are no other standard reference types today, but future ones (or custom ones) may not imply baggage propagation.

specification.md
```diff
@@ -0,0 +1,191 @@
...
+ [Span A···················································]
+   [Span B··············································]
+      [Span D··········································]
+    [Span C········································]
+         [Span E·······]        [Span F··] [Span G··] [Span H··]
+~~~
+
+Each **Span** encapsulates the following state:
+
+- A start timestamp
```
yurishkuro: + An operation name

specification.md
```diff
@@ -0,0 +1,191 @@
...
+
+#### Start a new `Span`
+
+Required parameters
+    - An **operation name**, a human-readable string which
+      concisely represents the work done by the Span (for example, an RPC
+      method name, a function name, or the name of a subtask or stage within a
+      larger computation). The operation name should be **the most general
+      string that identifies a (statistically) interesting class of `Span`
+      instances**. That is, `"get_user"` is better than `"get_user/314159"`.
```
yurishkuro: fwiw the current [spec](http://opentracing.io/documentation/pages/spec.html) had a table with 3 examples, maybe we can keep it.

yurishkuro:
@bensigelman do you think this should completely replace http://opentracing.io/documentation/pages/spec.html page? There's a very large overlap, but some explanation on the existing page are not included here, for example there was a page-long description of childOf/followsFrom references, which I think should be included in the Spec.

specification.md
```diff
@@ -0,0 +1,191 @@
...
+      [Span D··········································]
+    [Span C········································]
+         [Span E·······]        [Span F··] [Span G··] [Span H··]
+~~~
+
+Each **Span** encapsulates the following state:
+
+- A start timestamp
+- A finish timestamp
+- A set of zero or more key:value **Span Tags**. The keys must be strings. The
```
adriancole: are there no such constraints on "span logs"?

specification.md
```diff
@@ -0,0 +1,191 @@
...
+
+#### Log structured data
+
+Required parameters
+    - One or more key:value pairs, where the keys must be strings and the values may have any type at all. Some OpenTracing implementations may handle more (or more of) certain log values than others.
+
+Optional parameters
+    - An explicit timestamp. If specified, it must fall between the local start and finish time for the span.
+
+Note that the OpenTracing project documents certain **"standard log keys"** which have prescribed semantic meanings. TODO: add these log keys to a central .yaml file and link here.
```
adriancole: yeah especially as logging just a string is no longer in the api, this should be linked prior to merge.

specification.md
bensigelman: good point, added.

specification.md
bensigelman: good catch

specification.md
bensigelman: I had been planning to make this doc more formal, but upon reflection it's probably better to err on the side of over-explaining subtle things like this. Re-added.

specification.md
```diff
@@ -0,0 +1,191 @@
...
+**Returns** a `SpanContext` instance suitable for use as a **reference** when starting a new `Span` via the `Tracer`.
+
+#### Note: required **format**s for injection and extraction
+
+Both injection and extraction rely on an extensible **format** parameter that dictates the type of the associated "carrier" as well as how a `SpanContext` is encoded in that carrier. All of the following **format**s must be supported by all Tracer implementations.
+    - **Text Map**: an arbitrary string-to-string map with an unrestricted character set for both keys and values
+    - **HTTP Headers**: a string-to-string map with keys and values that are suitable for use in HTTP headers. That is, keys must not require strict casing and values must be URL-encoded
+    - **Binary**: a (single) arbitrary binary blob representing a `SpanContext`
+
+### `Span`
```
bensigelman: Good catch++

specification.md
bensigelman: Good call. I just did it and added to this PR.

specification.md
bensigelman: I don't think we should make that clarification until other reference types exist.

specification.md
```diff
@@ -0,0 +1,209 @@
...
+
+With the exception of the method to retrieve a `Span`'s `SpanContext`, no method may be called on a `Span` instance after it's finished.
+
+#### Set a `Span` tag
+
+Required parameters
+    - The tag key, which must be a string
+    - The tag value, which must be either a string, a boolean value, or a numeric type
+
+Note that the OpenTracing project documents certain **"standard tags"** which have prescribed semantic meanings. TODO: add these tags to a central .yaml file and link here.
```
yurishkuro: TODO can be fixed now

specification.md
yurishkuro: TODO can be fixed now

specification.md
```diff
@@ -0,0 +1,209 @@
...
+    - An **operation name**, a human-readable string which
+      concisely represents the work done by the Span (for example, an RPC
+      method name, a function name, or the name of a subtask or stage within a
+      larger computation). The operation name should be **the most general
+      string that identifies a (statistically) interesting class of `Span`
+      instances**. That is, `"get_user"` is better than `"get_user/314159"`.
+
+For example, here are potential **operation names** for a `Span` that gets hypothetical account information:
+
+| Operation Name | Guidance |
```
adriancole: nice

specification.md
```diff
@@ -0,0 +1,209 @@
...
+    - A **format** descriptor (typically but not necessarily a string constant) which tells the `Tracer` implementation how to decode `SpanContext` from the carrier parameter
+    - A **carrier**, whose type is dictated by the **format**. The `Tracer` implementation will decode the `SpanContext` from this carrier object according to **format**.
+
+**Returns** a `SpanContext` instance suitable for use as a **reference** when starting a new `Span` via the `Tracer`.
+
+#### Note: required **format**s for injection and extraction
+
+Both injection and extraction rely on an extensible **format** parameter that dictates the type of the associated "carrier" as well as how a `SpanContext` is encoded in that carrier. All of the following **format**s must be supported by all Tracer implementations.
+    - **Text Map**: an arbitrary string-to-string map with an unrestricted character set for both keys and values
+    - **HTTP Headers**: a string-to-string map with keys and values that are suitable for use in HTTP headers. That is, keys must not require strict casing and values must be URL-encoded
```
adriancole: http header values are documented as ISO-8859-1/ASCII octets or [mime encoded](https://tools.ietf.org/html/rfc2047) values. Since we aren't doing that, the "URL-encoded" part should be a link explaining why senders and receivers be looking at that vs what's normal.

adriancole: otherwise, it would be simple, as we'd just put a link to https://tools.ietf.org/html/rfc7230#section-3.2.4 referencing the standard encoding rules

specification.md
```diff
@@ -0,0 +1,209 @@
...
+
+With the exception of the method to retrieve the `Span`'s `SpanContext`, none of the below may be called after the `Span` is finished.
+
+#### Retrieve the `Span`s `SpanContext`
+
+There should be no parameters.
+
+**Returns** the `SpanContext` for the given `Span`. The returned value may be used even after the `Span` is finished.
+
+#### Overwrite the operation name
```
adriancole: interesting.

wu-sheng:
@bensigelman , what is the relation between specification.md and  data_conventions.yaml. It seems two separated files.

Maybe, should add a link in specification, guide reader to data_conventions ?

specification.md
bensigelman: yeah, this is something that comes up on the server side of various web frameworks where the Span is created before app-level data is necessarily available.

bensigelman:
@wu-sheng my intention with this repository was to make it more about data and specifications and less about "documentation" per se (which is the job of opentracing.io's repo). That said, the commit I'm about to make does include a cross-ref to the `.yaml` file from the spec file.

I do think the `.yaml` should be its own thing since various scripts could be built to process it and auto-generate standard tags in each language repository.

specification.md
bensigelman: @adriancole thanks for the IETF pointer. The trouble with HTTP headers is that they're so inconsistently implemented... i.e., being "standards-compliant" doesn't really count for much. I will add a link to the RFC 7230 and also a little more text to clarify.

wu-sheng:
@bensigelman , oh, I see.

bensigelman:
I'll merge this in 24-48h if there aren't further comments. Thanks!

wu-sheng:
@bensigelman , this will be a seperate page in spec doc? And should I add up a translation on this?

bensigelman:
@wu-sheng I was going to keep the existing `opentracing.io/spec` URL valid but make it much shorter... probably just a description of OpenTracing's master versioning scheme as described at #2, a link to the formal specification in this PR, and maybe some text about what the spec actually *means* (i.e., that it describes "pure" OpenTracing APIs and not their implementations from various vendors). Maybe there will be a diagram or something.

Regarding translation: I mean, it's great if you want to translate it, sure. It's a lot of ongoing work to maintain that but I'm certainly not complaining!

wu-sheng:
Yes, translation is always an on-going job, I assume. :) 
I just want to do it, on a checkpoint, or a frozen version.

PR #25, opened 2016-12-23
Add a website link on README

wu-sheng:
Advice by @ptone on https://github.com/opentracing/opentracing.io/issues/176. 

I think it's better to have it. @bensigelman

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # The OpenTracing Specification repository
 
-OpenTracing Specification Official Website: [http://opentracing.io/documentation/pages/spec](http://opentracing.io/documentation/pages/spec)
+OpenTracing Specification Official Website: [http://opentracing.io/spec](http://opentracing.io/spec).
 
 The github.com/opentracing/specification repository is a place to document (and discuss) the OpenTracing specification itself (independent of any particular language or platform).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,7 @@
 # The OpenTracing Specification repository
 
+OpenTracing Specification Official Website: [http://opentracing.io/documentation/pages/spec](http://opentracing.io/documentation/pages/spec)
+
```
yurishkuro: missing a period at the end.

yurishkuro: Also, this URL may be better: http://opentracing.io/spec

wu-sheng:
@yurishkuro Changed.

yurishkuro:
Thank, @wu-sheng

yurishkuro:
Sorry, I had to re-commit because of the bad commit message that gets added automatically. I changed the settings on the repo to always squash commits from now on.

wu-sheng:
: )  no need.  Do anything you need to be done.

README.md
```diff
@@ -1,5 +1,7 @@
 # The OpenTracing Specification repository
 
+OpenTracing Specification Official Website: [http://opentracing.io/spec](http://opentracing.io/spec).
```
bensigelman: @wu-sheng this is not the endgame I'd like here... let's close this and I will open an issue to resolve this properly from the other side. https://github.com/opentracing/opentracing.io/issues/177

bensigelman:
@yurishkuro this isn't really the solution I'd like, but I can just clean things up later.

wu-sheng:
What's the better solution?  Maybe I am not total understood.

And Happy holidays . @bensigelman @yurishkuro

PR #26, opened 2016-12-25
Document versioning policy

bensigelman:


bensigelman:
(writing down what we decided in Issue #2)

I'll merge this in a few days if there are no objections.

bensigelman:
I don't expect this to be controversial (was already discussed), so will merge. Thanks @yurishkuro :santa:

wu-sheng:
👏

Issue #30, opened 2017-01-09
Build OpenTracing reviewer team?


wu-sheng:
From @pavolloffay on git.

>Other thing I would find it useful if there were defined teams in ot-contrib(for reviews and ask in issues). People may wonder who should review their PR. It would be good if there were teams (one for spec and then per language probably)

Create team of spec and languages reviewers. More important, let people know who to ask and discuss. 
@adriancole @bensigelman @yurishkuro @michaelsembwever

pavolloffay:
I have created the same issue in `ot-contrib` : https://github.com/opentracing-contrib/meta/issues/2

wu-sheng:
From ben on gitter 

>wait a month or two so that some patterns emerge

adriancole:
>
> Create team of spec and languages reviewers. More important, let people
> know who to ask and discuss.
> @adriancole <https://github.com/adriancole> @bensigelman
> <https://github.com/bensigelman> @yurishkuro
> <https://github.com/yurishkuro> @michaelsembwever
> <https://github.com/michaelsembwever>
>
I can't commit to being on a review team. I'll be happy to help
infrequently, for example if we want to pow-wow ideas at meetups or such,
but I have to dial-down my responsibilities this year as I'm very near
burn-out.

wu-sheng:
Happy to be a member of spec and java-language reviewer team.

PR #37, opened 2017-02-08
Init the contributing.md for ot/spec.

wu-sheng:
Add a statement about `How about the contribution and governance model` from gitter. Although the OpenTracing Team Members List is out-of-date.

>at this point I have a reasonable sense of which people typically care about which sorts of things
maybe it’s time to do something more formal… /shrug, open to opinions.

I think this is the first step. And we should set a formal **OpenTracing Team Members List**.

If there is any English-language issue, please help me fix them. 😝

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+CONTRIBUTING
+============
+
+Contributions are always welcome, no matter how large or small. Before contributing, you should know, this repository is a place to document Official OpenTracing Specification, so, OpenTracing team will **treat all Contributions very carefully**.
+
+### Understand Specification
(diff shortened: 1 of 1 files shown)


bensigelman:
Thanks for kicking this off... rather than nit-picking your English, I will just write a parallel version and send it out as a separate PR.

wu-sheng:
@bensigelman , It's  up to you.😎

wu-sheng:
These were two issues https://github.com/opentracing-contrib/meta/issues/2 and https://github.com/opentracing/specification/issues/30, which are related to the contributing and formal members ( or reviewer team ).

I think we should deal with these things. Last time, I and @pavolloffay bring this up, is one month ago. :p 

@bensigelman @yurishkuro @michaelsembwever thoughts?

bensigelman:
See #39... they are not exactly the same thing, though.

PR #44, opened 2017-02-23
Add a db tag guidance.

wu-sheng:
Right now, include two demonstrations.
- Java, `Spring JdbcTemplate -> MySQL`
- Python, `redis-py  -> Redis`

### DIFF
--- a/guidance/database/tags.md
+++ b/guidance/database/tags.md
@@ -1,0 +1,60 @@
+# Database tags guidance
+
+This guidance shows how to tag span, when you trace a Database access.
+
+Here are some demonstrations.
+
+### Trace a MySQL query statement
+Background story:
+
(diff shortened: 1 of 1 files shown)


guidance/database/tags.md
```diff
@@ -0,0 +1,62 @@
+# Database tags guidance
```
bhs: I like the level of detail here, but I think we should wait to merge until the `data_conventions.yaml` becomes an .md file. I will try (?) to get that done today, assuming I can find the time!

wu-sheng: I will wait for you. It's better, that you write the `data_conventions.md`.

bhs: (https://github.com/opentracing/specification/pull/45)

wu-sheng:
Closing it. Include in #45 .

PR #45, opened 2017-02-25
Move most of the semantic data conventions to .md

### LINKED ISSUE #41: Rework the formal key and field guidance as an .md file
The .yaml file was a nice idea, but since there turns out to be language-specific logic (e.g., for errors). I think we should create an .md file to document these tags/fields, then make the yaml file really just the data (i.e., without comments and guidance on that front).

bhs:
... there is still a .yaml file, but it's now just machine-oriented (all comments and examples are now part of an .md file instead).

Closes #41, obviates (??) #44.


semantic_conventions.md
```diff
@@ -0,0 +1,82 @@
+# Semantic Conventions
+
+The [OpenTracing Specification](https://github.com/opentracing/specification/blob/master/specification.md) describes the overarching language-neutral data model and API guidelines for OpenTracing. That data model includes the related concepts of **Span Tags** and **(structured) Log Fields**; though these terms are defined in the specification, there is no guidance there about standard Span tags or logging keys.
```
yurishkuro: relative URL `./specification.md` would be more portable

semantic_conventions.md
```diff
@@ -0,0 +1,82 @@
...
+| `http.status_code` | integer | HTTP response status code for the associated Span. E.g., 200, 503, 404 |
+| `http.url` | string | URL of the request being handled in this segment of the trace, in standard URI format. E.g., '"https://domain.net/path/to?resource=here"' |
+| `peer.hostname` | string | Remote hostname. E.g., "opentracing.io", "internal.dns.name" |
+| `peer.ipv4` | string | Remote IPv4 address as a .-separated tuple. E.g., "127.0.0.1" |
+| `peer.ipv6` | string | Remote IPv6 address as a string of colon-separated 4-char hex tuples. E.g., "2001:0db8:85a3:0000:0000:8a2e:0370:7334" |
+| `peer.port` | integer | Remote port. E.g., 80 |
+| `peer.service` | string | Remote service name (for some unspecified definition of "service"). E.g., "elasticsearch", "a_custom_microservice", "memcache" |
+| `sampling.priority` | integer | If greater than 0, a hint to the Tracer to do its best to capture the trace. If 0, a hint to the trace to not-capture the trace. If absent, the Tracer should use its default sampling mechanism. |
+| `span.kind` | string | Either "client" or "server" for the appropriate roles in an RPC. E.g., "client", "server" |
```
yurishkuro: The original idea of a yaml file was to be the source from which the constants in various APIs, as well as this table, could be generated. So I rather liked the previous yaml versions which specified name, type, and description, while the larger comments can be here.

semantic_conventions.md
bhs: Done

semantic_conventions.md
bhs: My thought was that, if/when the generated code actually happens, it would have a big block comment pointing to the `.md` file for semantic details, esp about combinations of tags/fields.

yurishkuro:
> My thought was that, if/when the generated code actually happens, it would have a big block comment pointing to the .md file for semantic details, esp about combinations of tags/fields.

Yes, that's fine, but all constants would have no comments (go linter will be sad). But we can revisit then.

semantic_conventions.yaml
```diff
@@ -0,0 +1,32 @@
+# This file describes standards and conventions for Span tags.
```
yurishkuro: tags and logs

bhs:
@yurishkuro 

> Yes, that's fine, but all constants would have no comments (go linter will be sad). But we can revisit then.

Yeah, exactly. I was starting to feel like the .yaml was a "pre-optimization", or at least a bit ahead of its time... if we're going to stick all of the comments in there and do codegen, we ought to be codegen-ing the `.md` file, too. Just seems like the need isn't there yet.

semantic_conventions.yaml
bhs: Fixed, good eye...

bhs:
(I also added a few backlink fixes)

README.md
```diff
@@ -10,5 +10,5 @@ Historically this activity took place on the docs repo ([github.com/opentracing/
 
 There are two important files in this repository:
 
-1. [`specification.md`](https://github.com/opentracing/specification/blob/master/specification.md), a versioned description of the current pan-language OpenTracing standard
-1. [`data_conventions.yaml`](https://github.com/opentracing/specification/blob/master/data_conventions.yaml), a structured YAML file describing standard Span tags and Span log keys
+1. [`specification.md`](./specification.md), a versioned description of the current pan-language OpenTracing standard
+1. [`semantic_conventions.md`](./semantic_conventions.yaml), a document describing conventional Span tags and log keys for common semantic scenarios
```
yurishkuro: did you mean the link to go do .md as well?

bhs: yep

semantic_conventions.md
```diff
@@ -0,0 +1,82 @@
...
+
+| Span tag name | Type | Notes and examples |
+|:--------------|:-----|:-------------------|
+| `component` | string  | The software package, framework, library, or module that generated the associated Span. E.g., `"grpc"`, `"django"`, `"JDBI"`. |
+| `db.instance` | string | Database instance name. E.g., In java, if the jdbc.url=`"jdbc:mysql://127.0.0.1:3306/customers"`, the instance name is `"customers"`. |
+| `db.statement` | string | A database statement for the given database type. E.g., for `db.type="SQL"`, `"SELECT * FROM wuser_table"`; for `db.type="redis"`, `"SET mykey 'WuValue'"`. |
+| `db.type` | string | Database type. For any SQL database, `"sql"`. For others, the lower-case database category, e.g. `"cassandra"`, `"hbase"`, or `"redis"`. |
+| `db.user` | string | Username for accessing database. E.g., `"readonly_user"` or `"reporting_user"` |
+| `error` | bool | `true` if and only if the associated Span is in an error state |
```
yurishkuro: would this be more concrete: "if the application considers the operation represented by the Span to have failed"

semantic_conventions.md
```diff
@@ -0,0 +1,82 @@
...
+| `http.status_code` | integer | HTTP response status code for the associated Span. E.g., 200, 503, 404 |
+| `http.url` | string | URL of the request being handled in this segment of the trace, in standard URI format. E.g., `"https://domain.net/path/to?resource=here"` |
+| `peer.hostname` | string | Remote hostname. E.g., `"opentracing.io"`, `"internal.dns.name"` |
+| `peer.ipv4` | string | Remote IPv4 address as a `.`-separated tuple. E.g., `"127.0.0.1"` |
+| `peer.ipv6` | string | Remote IPv6 address as a string of colon-separated 4-char hex tuples. E.g., `"2001:0db8:85a3:0000:0000:8a2e:0370:7334"` |
+| `peer.port` | integer | Remote port. E.g., `80` |
+| `peer.service` | string | Remote service name (for some unspecified definition of `"service"`). E.g., `"elasticsearch"`, `"a_custom_microservice"`, `"memcache"` |
+| `sampling.priority` | integer | If greater than 0, a hint to the Tracer to do its best to capture the trace. If 0, a hint to the trace to not-capture the trace. If absent, the Tracer should use its default sampling mechanism. |
+| `span.kind` | string | Either `"client"` or `"server"` for the appropriate roles in an RPC. E.g., `"client"`, `"server"` |
```
yurishkuro: can remove `E.g., "client", "server"`

semantic_conventions.yaml
```diff
@@ -0,0 +1,32 @@
+# This file describes standards and conventions for Span tags and logs.
```
yurishkuro: there are no links to this file - per earlier discussion maybe we should delete it altogether, so that .md file is the only source of truth?

bhs: I can't remember where (it wasn't on this PR), but there were some people who wanted it to continue to exist within the past week.

bhs:
I'll give others 48h to raise objections prior to merging.

PR #46, opened 2017-02-26
Add clarification about OT scope and Span logging

bhs:
At the distributed tracing workshop, the intended scope and purpose of logging (and, in some cases, OpenTracing itself) was not as clear as I would have hoped. This PR doesn't "fix" that communication problem, but hopefully makes things incrementally better.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -10,6 +10,12 @@ This is the "formal" OpenTracing semantic specification. Since OpenTracing must 
 
 The OpenTracing specification uses a `Major.Minor` version number but has no `.Patch` component. The major version increments when backwards-incompatible changes are made to the specification. The minor version increments for non-breaking changes like the introduction of new standard tags, log fields, or SpanContext reference types. (You can read more about the motivation for this versioning scheme at Issue [specification#2](https://github.com/opentracing/specification/issues/2#issuecomment-261740811))
 
+## The Big Picture: OpenTracing's Scope
+
+OpenTracing's core specification (i.e., this document) is intentionally agnostic about the specifics of particular downstream tracing or monitoring systems. This is because **OpenTracing exists to describe the semantics of transactions in distributed systems.** Describing those transactions should not be influenced by how — or how not — any particular backend likes to process or represent data. For instance, detailed OpenTracing instrumentation can be used to simply measure latencies and apply tags in a timeseries monitoring system (e.g., prometheus); or Span start+finish times along with Span logs may be redirected to a central logging service (e.g., Kibana).
+
+As such, the OpenTracing specification and [data modelling conventions](./data_conventions.md) have a wider scope than some tracing systems, "and that's okay." If certain semantic behavior is out-of-scope for a particular tracing or monitoring system, said system can summarize or simply ignore the respective data flowing from OpenTracing instrumentation.
+
 ## The OpenTracing Data Model
 
 **Traces** in OpenTracing are defined implicitly by their **Spans**. In
@@ -218,6 +224,14 @@ Optional parameters
 - An explicit timestamp. If specified, it must fall between the local start and finish time for the span.
 
 Note that the OpenTracing project documents certain **["standard log keys"](./semantic_conventions.md)** which have prescribed semantic meanings.
+
+##### An aside: "Logging" in general, and what it means in OpenTracing
+
+"Logging" is an overloaded term in our industry; one could reasonably argue that all tracing is just a particularly organized form of logging. OpenTracing "logs" are really just key:value maps that describe a particular moment within the context of a Span.
+
+It is not forbidden but also not particularly wise to redirect general-purpose process-level logging into OpenTracing. Some of those logs may be relevant to transaction traces, but some may also pertain to process-global state which has no direct relationship to a transaction trace.
+
+The granularity of Span logs is intended to be finer than typical "info"-style logging in process-level logging frameworks. Since tracing systems usually have a smart, all-or-nothing per-trace sampling mechanism, the verbosity within a single trace can be higher than what would be appropriate for a process as a whole — especially when that process contends with high concurrency. |
 
 #### Set a **baggage** item
 
--- a/specification.md
+++ b/specification.md
@@ -12,7 +12,7 @@ The OpenTracing specification uses a `Major.Minor` version number but has no `.P
 
 ## The Big Picture: OpenTracing's Scope
 
-OpenTracing's core specification (i.e., this document) is intentionally agnostic about the specifics of particular downstream tracing or monitoring systems. This is because **OpenTracing exists to describe the semantics of transactions in distributed systems.** Describing those transactions should not be influenced by how — or how not — any particular backend likes to process or represent data. For instance, detailed OpenTracing instrumentation can be used to simply measure latencies and apply tags in a timeseries monitoring system (e.g., prometheus); or Span start+finish times along with Span logs may be redirected to a central logging service (e.g., Kibana).
+OpenTracing's core specification (i.e., this document) is intentionally agnostic about the specifics of particular downstream tracing or monitoring systems. This is because **OpenTracing exists to describe the semantics of transactions in distributed systems.** Describing those transactions should not be influenced by how — or how not — any particular backend likes to process or represent data. For instance, detailed OpenTracing instrumentation can be used to simply measure latencies and apply tags in a timeseries monitoring system (e.g., Prometheus); or Span start+finish times along with Span logs may be redirected to a central logging service (e.g., Kibana).
 
 As such, the OpenTracing specification and [data modelling conventions](./data_conventions.md) have a wider scope than some tracing systems, "and that's okay." If certain semantic behavior is out-of-scope for a particular tracing or monitoring system, said system can summarize or simply ignore the respective data flowing from OpenTracing instrumentation.
 
@@ -229,9 +229,9 @@ Note that the OpenTracing project documents certain **["standard log keys"](./se
 
 "Logging" is an overloaded term in our industry; one could reasonably argue that all tracing is just a particularly organized form of logging. OpenTracing "logs" are really just key:value maps that describe a particular moment within the context of a Span.
 
-It is not forbidden but also not particularly wise to redirect general-purpose process-level logging into OpenTracing. Some of those logs may be relevant to transaction traces, but some may also pertain to process-global state which has no direct relationship to a transaction trace.
+While it's possible to redirect general-purpose process-level logging into OpenTracing, doing so requires care. For instance, logging statements that aren't anchored in specific transactions or traces may not make sense within a tracing system. That said, in enviroments where an overwhelming fraction of conventional logging statements already refer to distributed transactions, `tee`ing that logging data into OpenTracing is reasonable and often beneficial.
 
-The granularity of Span logs is intended to be finer than typical "info"-style logging in process-level logging frameworks. Since tracing systems usually have a smart, all-or-nothing per-trace sampling mechanism, the verbosity within a single trace can be higher than what would be appropriate for a process as a whole — especially when that process contends with high concurrency. |
+The granularity of Span logs is intended to be finer than typical "info"-style logging in process-level logging frameworks. Since tracing systems usually have a smart, all-or-nothing per-trace sampling mechanism, the verbosity within a single trace can be higher than what would be appropriate for a process as a whole — especially when that process contends with high concurrency.
 
 #### Set a **baggage** item
 
--- a/semantic_conventions.md
+++ b/semantic_conventions.md
@@ -63,9 +63,15 @@ The following Span tags combine to model database calls:
 - `peer.hostname`, `peer.ipv4`, `peer.ipv6`, `peer.port`, `peer.service`: optional tags that describe the database peer
 - `span.kind`: `"client"`
 
-### Captured errors
+### Span and log errors
 
-Errors may be described by OpenTracing in different ways, largely depending on the language. Some of these descriptive fields are specific to errors; others are not (e.g., the `event` or `message` fields).
+It is important to distinguish between **error Spans** and **errors logged during Span execution**.
+
+Every Span either finishes in an error state or does not: the `"error"=true` tag distinguishes between those two cases. (If the `"error"` tag is missing altogether, that implies `"error"=false`) Tools that consume OpenTracing instrumentation should not need to consider any other information to determine whether a Span is in an error state.
+
+#### Logged application-level errors
+
+It can also be useful to record application-level errors that crop up during a Span's lifetime. For those situations, Span logs are more appropriate since errors have a specific timestamp (and Spans in general represent a time interval, not a specific moment). Logged errors may be described by OpenTracing in different ways, largely depending on the language. Some of these descriptive fields are specific to errors; others are not (e.g., the `event` or `message` fields).
 
 For languages where an error object encapsulates a stack trace and type information, log the following fields:
 
@@ -80,3 +86,5 @@ For other languages, or when above is not feasible:
 - error.kind=`"..."` (optional)
 
 This scheme allows Tracer implementations to extract what information they need from the actual error object when it's available.
+
+**Note:** a Span may be in an error state (i.e., have an `"error"=true` tag) and have no error *logs*, and vice versa.


specification.md
```diff
@@ -219,6 +225,14 @@ Optional parameters
...
 
+##### An aside: "Logging" in general, and what it means in OpenTracing
+
+"Logging" is an overloaded term in our industry; one could reasonably argue that all tracing is just a particularly organized form of logging. OpenTracing "logs" are really just key:value maps that describe a particular moment within the context of a Span.
+
+It is not forbidden but also not particularly wise to redirect general-purpose process-level logging into OpenTracing. Some of those logs may be relevant to transaction traces, but some may also pertain to process-global state which has no direct relationship to a transaction trace.
+
+The granularity of Span logs is intended to be finer than typical "info"-style logging in process-level logging frameworks. Since tracing systems usually have a smart, all-or-nothing per-trace sampling mechanism, the verbosity within a single trace can be higher than what would be appropriate for a process as a whole — especially when that process contends with high concurrency. |
+
```
yurishkuro: odd `|` at the end

specification.md
```diff
@@ -219,6 +225,14 @@ Optional parameters
 
 Note that the OpenTracing project documents certain **["standard log keys"](https://github.com/opentracing/specification/blob/master/data_conventions.yaml)** which have prescribed semantic meanings.
 
+##### An aside: "Logging" in general, and what it means in OpenTracing
+
+"Logging" is an overloaded term in our industry; one could reasonably argue that all tracing is just a particularly organized form of logging. OpenTracing "logs" are really just key:value maps that describe a particular moment within the context of a Span.
+
+It is not forbidden but also not particularly wise to redirect general-purpose process-level logging into OpenTracing. Some of those logs may be relevant to transaction traces, but some may also pertain to process-global state which has no direct relationship to a transaction trace.
```
yurishkuro: "but also not particularly wise" - maybe we can avoid giving an opinion here. It actually depends on the use cases and how it's done, and in certain cases it could be very beneficial to fork (rather than redirect) request-scoped logs into OT.

bhs: I rewrote this paragraph...

specification.md
bhs: ah yes -- a vestige of a previous attempt to make this aside part of a "sidebar" (attempt was not successful!)

bhs:
I'll merge this in ≥ 48h if there are no objections.

adriancole:
TL;DR; this clarifies a view I believe shared between @bhs and @yurishkuro, but is somewhat different to how most apply tracing apis. While the text says all of this is optional, it is OT taking a stance: OT already accepted apis that aren't largely supported, and reinforced their usage by pinning other standards to them (Ex. error logging) and proliferating it in instrumentation practice. This doc moreover also clarifies that fine grained logging is not only supported, but expected.

That's fine, but both of these lead to further lock-in until/unless other systems decide to follow suit. For example, I doubt users will want to use a system that cannot support something as basic as the OT "error" standard (in OT errors are KV logging). This may lead to different standards even if they are layered further on OT, for example a "simplified" tracing api which has been discussed before.

There are two points elaborated on below:
* documentation is how people can understand whether or not features exist
* logging (annotations in zipkin) are intended to be used at a finer grain than "info" logging

### On using documentation to indicate if any non-structural features are supported.

The first point is basically the definition of a leaky abstraction. This is not intended to be an debasing comment, just a reality. As mentioned in the workshop, currently tracing apis are at the very depths of the code, far away from users. Layering on top occurs sometimes twice, ex. to instrument a framework, then to pin instrumentation to a configuration tool or framework.

Let's say someone wrote an OpenTracing instrumentation assuming KV logging is supported. In fact, recent java instrumentation written by @pavolloffay and also the definition of how to [write an error](https://github.com/opentracing/specification/pull/35/files) make an assumption that KV is supported. A user of instrumentation will likely not read this specification, and instead win by the understood charter of "instrument once". When they turn on this, they'll notice that only systems that support indexed nested KV logging will have data that can be searchable, even for features as primary as errors. While there are some commercial APMs which might support this, it is unclear which as there's no feature registry. At the moment, in open source at least, this maybe limits to only jaeger? I think users of an open abstraction api would be surprised that the most core features including errors and how instrumentation are written limit so much.

The long story short is that for a user to understand what will work, they will need to look at each instrumentation project, see which features are used by it, then compare that against a currently non-existent feature matrix of backend systems. The simple choice would likely be to use jaeger I guess.. but that would also be invalidated if a new api is made regardless of jaeger, and jaeger chooses not to support it.

### On assuming logs (annotations in zipkin) are intended to be used at a finer granularity than "info"

This is somewhat related to the above, as the change here sets expectations to something that disfavors brown field. During the tracing workshop, the clear majority present agreed that status quo tracing is less granular to logging rather than more granular, save perhaps google internally. @peterbourgon made this nice diagram that largely fits existing practice, which is that tracing is at least status quo at request granularity:

screen shot 2017-02-27 at 10 17 43 am

Specifically to logging or timeline annotations, this fits as well. For example, finagle (via zipkin) prefers fixed-length spans and very judiciously uses its `Trace.record` api, resulting in relatively fixed-size spans (cc @mosesn). This is the same in htrace: in HBase instrumentation, there's only several usages of `span.addTimelineAnnotation` (cc @cmccabe). Public cloud providers including google and amazon don't even expose such an api (timestamped values) (cc @abhiksingh @mtwo). I'd inventory other APMs, but whether or not they use OpenTracing, and the degree each feature is supported is generally unknown.

Besides Jaeger (cc @yurishkuro), which is green-field OpenTracing, the only systems I'm aware of in open source that might assume <log.info practice might be logging systems. Ex maybe wingtips (cc @nicmunroe)

As noted in the workshop, taking the stance that spans can be used as a replacement for logging (even if noted to do so with care) is surprising considering status quo. The suggestion that use of a span's logging (timestamp annotation) api is likely to be more volume than log.info is even more surprising. As also noted in the workshop, this has backend considerations which even if don't affect some (ex @abhiksingh mentioning you could theoretically put a GiB in a span), it certainly impacts the design of many.

### Overall opinion

@peterbourgon raised a question in the workshop to the effect of whether OpenTracing is choosing to lead (create a new api) or follow (existing practice). Both of the clarifications suggest quite clearly that OpenTracing is choosing to lead. Even if there's a statement saying "see docs", this is impractical for any realistic interop: other specs like error, and routine instrumentation churns out under the assumption KV logging is supported, making it a primary feature. Reinforcing comments including that more granular logging is expected rather than the exception is also a leadership position in the realm of a minimum viable feature set.

It is no problem that OT leads, just that it goes farther against the "instrument once" charter, and it also selects green field implementations implicitly, including almost all known tracing systems. This may ironically lead to lock-in, as it is unrealistic to expect casual users to do a depth-first comparison of instrumentation vs feature sets when we know neither exist today, and there are no interop tests even being planned to produce them.

In 4 years of Zipkin, there has been not even a single user request for KV logging until one a few months ago which was in response to OT defining the api. Many of our sites are short staffed, where 10% or 50% of one person supports all of tracing including api questions. In Twitter, it was usually less than one person. Adding fine-grained logging support to zipkin has impact all over the place including kafka message sizes and storage, as well support implications of a larger scope. This would come at the cost of what users actually ask for, and raise the requirements of site administration. Companies with larger staff, who implement tracing with logging, or who use a commercial APM can likely afford the implicit direction of OT. However, the lightly staffed and/or brown field ecosystem may be left with more work or responsibility as a result of the same. That, or they opt-out of OT to avoid it.

Finally, the most important part is that even though this change in text says everything is optional, in practice there's not yet governance to ensure anything is optional (cc @basvanbeek who's been tapped to help). If there was, the "optional parts" would be at least noted in READMEs etc of the fast landing new instrumentation projects. Until "instrument once" is realistic in OT (ie ensured with at least docs if not api support), I would hesitate proliferating that message as by default it is misleading.

yurishkuro:
> recent java instrumentation written by @pavolloffay and also the definition of how to write an error make an assumption that KV is supported.

The minimum recommended level of marking errors (although I don't think it's explicitly written anywhere) is `error: true` tag , which seems to be supported by most implementations. I think it makes sense to state this more explicitly in the spec - (cc @bhs) - as part of describing "the semantics of transactions in distributed systems". Logging of the error is to provide additional (_optional_) information beyond the Boolean marker on the span. If instrumentation only logs the errors, I think it's incomplete. The fact that `log(event: 'error')` is called should not be interpreted by the tracer as "span has failed".

adriancole:
@yurishkuro thanks for the clarification. I opened #47 for follow-up on required vs nice-to-have on error support

beberlei:
I think the logging part would fall much easier in place, if we had access to the current active span, its id and trace id. Then hooking this information into a logging framework is trivial. Logging frameworks are well understood and having the IDs would allow correlation, which is also easily understood.

As I said in the workshop, logging and tracing usually follows different sampling rates and the backends, processing and storage for both are very different. An efficient, large-scale solution would take this into account. But that means that at the Tracer level, users need to aggregate this into different tools, which is not a use-case that is documented at all. Thus users try to solve everything with just one tool (Zipkin), and the problems begin.

peterbourgon:
@beberlei is it correct that you're suggesting something like this architecture?

```
+------------------+
| Application code |
+------------------+
  | OpenTracing
  v
+-----------------------------------------+
| Implementation library e.g. Zipkin      |
+-----------------------------------------+
  | Log events             | Trace data
  v                        v
+--------------------+   +----------------+
| Log aggregator     |   | Span collector |
| e.g. Elasticsearch |   | e.g. Cassandra |
+--------------------+   +----------------+
```

If so, that seems like quite the scope creep for OpenTracing implementations!

beberlei:
@peterbourgon probably yes, i would visualize it this way:

```
+------------------+
| Application code |
+------------------+
  | OpenTracing
  v
+-----------------------------------------+
| Tracer delegator                        |
+-----------------------------------------+
  | Log events             | Trace data
  v                        v
+--------------------+   +----------------+
| Log Tracer         |   | Span Tracer    |
| Full Sampling      |   | RandomSampling |
+--------------------+   +----------------+
```

This makes sense and is necessary, if you use OpenTracing to do everything (logging, tracing, metrics). I don't think thats necessarily a good idea though, logging has so well defined APIs that I would probably prefer using them and correlate through IDs that i get from the active tracer/span.

yurishkuro:
From experience of deploying tracing at a large org, I don't see the approach of requiring developers to log through OT API getting a lot of traction. Therefore, our approach was to bind standard logging APIs to the current span behind the scene ([example](https://github.com/uber/jaeger/blob/master/examples/hotrod/pkg/log/factory.go)).

beberlei:
@yurishkuro Yes thats what I think, i think we should make clear that this is the much better approach, as far as I understand Adrians feedback from the Zipkin community, users of opentracing confuse logKV with a replacement with logging, because of the instrument once, use everywhere approach and then push everything into Zipkin. We should document that this is not what logKV is for. I feel code samples are worth more than 1000 words here.

yurishkuro:
> i think we should make clear that this is the much better approach

Well, except that I am reluctant to start labelling different approaches as better/worse, I feel it's a premature generalization.  Conceptually, there are two types of events people log, request-scoped and not request-scoped.  The common logging APIs do not distinguish them. If you look at my [example](https://github.com/uber/jaeger/blob/master/examples/hotrod/pkg/log/factory.go), the logging factory does not actually have any log methods, it requires the developer to make an explicit choice and get the logger via `Bg()` or `For(ctx)` (request scoped). Once the developer makes this choice, whichever API is returned from `For(ctx)` is largely irrelevant - it could be OT, but for convenience I chose to return the same logging API as in `Bg()`. So make a choice (the stick), get familiar API back (the carrot). At that point whether the resulting logs end up in tracing or logging backend (or both) is a matter of configuration.

So as far as the spec goes, I agree that it can make suggestions (code examples) and show alternative ways of dealing with request-scoped logging, without assigning value to them.

NB: the above is less relevant for languages like Java where the request-scoped logs can be detected automatically (via access to the active span in a thread local).

bhs:
@adriancole great summary and survey of how various tracing systems treat "logs" (for lack of a better word), thank you.

Gosh, so much to say. I will try to be concise.

### Regarding whether OpenTracing is "leading" or "following"

IMO, "false dichotomy." From this actual PR:

> OpenTracing's core specification (i.e., this document) is intentionally agnostic about the specifics of particular downstream tracing or monitoring systems. This is because **OpenTracing exists to describe the semantics of transactions in distributed systems.**

The only reason "logging" (if that's the right word) made it into the spec is because there are important moments in the life of distributed transactions that don't warrant their own Span. Thus, some finer-grained abstraction wanted to exist to describe those.

### Regarding whether KV is the right way to represent logging data

It's not obvious. It's easy to make arguments for and against.

My favorite argument against KV logging is that it's more complicated than simple string logging.

My favorite argument for KV logging is – at the risk of sounding like a broken record – it is the most direct way to model the semantics of transactions in distributed systems... and there's always the fallback of dumb/trivial JSONification (or similar) for unrecognized combinations of keys, so it doesn't place an undue burden on tracing systems that don't want to write code for each such combination.

### Regarding the "intended" granularity of logs in OT

I'm convinced that this should be non-specified. It makes assumptions about what's normal outside of OT, and that varies too much to be a useful baseline.

### Regarding `tee`-ing process-level logging

I think it's fine/desirable to tee process-level logging *when it's about specific transactions*. I don't like applying non-transaction-scoped logs to OpenTracing, and for the same broken-record reason as cited above.

That said, I also agree wholeheartedly with @beberlei's comments about how a generally accessible Span is basically a requirement for any such integration.

### Regarding errors

@yurishkuro pointed out that the `error=true` tag is the ground truth for errors in OT... the log formatting thing is just an attempt to specific how exceptions/etc ought to be handled. I agree that this needs to be clearer.

### In closing...

I'd like to see what other thoughts come up on this thread. Ideally the actual .md files in this PR can be improved to address some of the subtleties brought up here and at the workshop.

#23 also plays an important role in all of this... so I'm hoping to see substantial/tangible progress there in the next week as well.

bhs:
... I've pushed a commit that tries to further clarify the distinction between "error Spans" and logged application exceptions/errors.

yurishkuro:
@opentracing/otsc @opentracing/otiab please vote

wu-sheng:
+1 on this clarification.

specification.md
```diff
@@ -219,6 +225,14 @@ Optional parameters
 
 Note that the OpenTracing project documents certain **["standard log keys"](./semantic_conventions.md)** which have prescribed semantic meanings.
 
+##### An aside: "Logging" in general, and what it means in OpenTracing
+
+"Logging" is an overloaded term in our industry; one could reasonably argue that all tracing is just a particularly organized form of logging. OpenTracing "logs" are really just key:value maps that describe a particular moment within the context of a Span.
+
+While it's possible to redirect general-purpose process-level logging into OpenTracing, doing so requires care. For instance, logging statements that aren't anchored in specific transactions or traces may not make sense within a tracing system. That said, in enviroments where an overwhelming fraction of conventional logging statements already refer to distributed transactions, `tee`ing that logging data into OpenTracing is reasonable and often beneficial.
```
mansu: I think adding a warning/note here that says something along the lines of "adding too many logs to a span will increase the size of the span which in turn can have a negative performance impact on application performance and trace pipeline" would be very useful. 

I have seen this issue a few times now in our production environment. Further, this is true for all tracing backends currently, so adding a warning in these docs is appropriate.

mansu:
Added a minor comment but +1 otherwise.

adriancole:
Before encouraging something that many even here are uncomfortable with by saying calling it "often beneficial" .. maybe have some major vendors with a lot of experience validate this conjecture? Maybe dynatrace app dynamics, new relic?

adriancole:
I havent done an exhaustive search, but I cant find a major vendor sdk that conflates logging api with span api (even if there are both apis present in their greater APM sdk). Maybe someone here knows better, but with neither major players being involved in this spec, and this conflation largely absent from their impls, it seems strange to call such a practice "often beneficial".

I'd be interested to know of the tracing api in my TV decided to use this approach because it was beneficial despite all of the complaints and rationale discussed. Wouldnt it be nice to hear from established players? .. to know the difference between something they dont do because they havent found it beneficial yet, or if they choose not to because they tried and found it not beneficial, or that there is an alternate way (like normal log appender etc) which works better.

Even if something is often beneficial.. why? Is it because it is beneficial to a saas that has no log integration mechanism otherwise? Is it beneficial holding something constant such as advanced log search and bulk transfer reporting capabilities?

If open tracing had a separate logging api, like census do, how much value is lost? Is it really required to have these apis have more surface area to reach the same end goal? What does it cost to do recommend this by saying it is often beneficial? Ex does this continue to dig into the static initialization area (because logs are forwarded to span api), does it cost more to site admins who now have more data to filter, changing their tracing role to a higher percentage of hours. Does it cost more to be a tracing supplier? Ex do others now need to consider something besides cassandra because routine logs and the requirements around them outgrow the feature set. 

If this practice is really that encouraged, it is inflating yet more responsibility on others. Being more direct about costs, more direct about others not being on board with this, and more direct about existing practice are in my opinion ways to turn this around.

adriancole:
@jmkaldor I noticed you are now here, which sounds like progress. I presume FB are OpenTracing since you are on the board. Does FB find it often beneficial to route logs to the span api? If so, why? If not, why not?

bhs:
@adriancole thank you for the note as always.

> Even if something is often beneficial.. why?

For background and to emphasize that "logging-in-tracing" is not a new idea, I will start with the dapper paper... from the section where the Span data model is actually introduced (emphasis added):

> **2.1 Trace trees and spans**
> In a Dapper trace tree, the tree nodes are basic units of work which we refer to as spans. The edges indicate a casual relationship between a span and its parent span.  Independent of its place in a larger trace tree, though, **a span is also a simple log of timestamped records** which encode the span’s start and end time, any RPC timing data, and zero or more application-specific annotations as discussed in Section 2.3.

It's obvious, but it's worth mentioning that Zipkin also has this "simple log" (expressed as a timestamped-span-annotation) concept... so what I'm essentially arguing for is already supported. To be crystal-clear, I am not suggesting that a/any tracing system replaces a conventional centralized logging system. But those centralized logging systems are having ROI trouble as they try to describe distributed transactions... I can speak to that a bit, maybe it will help explain?

### Troubles with centralized logging and microservices

Why would anyone want to "mix metaphors" and send something that looks like ELK/Splunk data into "a tracing system"? Let me try to briefly explain the problem as it's been expressed to me (over and over again):

image

Tracing systems are relevant here because they have an inbuilt sampling mechanism that's well-suited to dealing with fine-grained **transactional** logging / "timestamped annotation" data. I am not expecting or proposing that tracing systems build out a fully-formed log-analysis feature set, but I *am* expecting (more like "knowing") that more and more organizations are having to literally turn off swaths of logging throughout their stack because the per-transaction-per-microservice data is so voluminous that they can't afford to stick it in ELK/splunk anymore.

BTW, I imagine it's irritating for me to say "over and over again" but then not cite things... this is mostly because of formal and informal NDAs. That said, Paul Fisher gave an excellent public talk at Monitorama digging into centralized logging at Lyft which also covers a lot of the problems with logging in a true microservice architecture (among many other things)... here's a screenshot:

image

Here's a link to the talk video: https://vimeo.com/221066726

### TLDR

It's about sampling and displaying **logs that are too detailed for the (cost-constrained) logging system** in the context of traces. It is not about providing a full log-analysis feature set, which I agree is an unreasonable expectation.

----------

PS: re facebook at OT...

> @jmkaldor I noticed you are now here, which sounds like progress. I presume FB are OpenTracing since you are on the board.

I made a deliberate effort to get people onto the OTIAB who were not necessarily using OT per se... this was specifically to avoid echo-chamber thinking, etc. To the best of my knowledge, FB is not using OT (or at least not in earnest), though maybe I'm wrong. :) I would be interested to hear Jonathan's take, so thank you for cc'ing him directly.

jmkaldor:
Uhh, hi all! I am still catching up on, well, everything, and I also tend to be super verbose, so I'll (try to) keep it short.

As @bhs notes, FB is currently not using OT — we're using an in-house tracing system we've been developing over the past few years, and we've been talking off and on with @yurishkuro and @bhs for the past 7-or-so months to compare notes. Our hope is to talk more with others in the tracing community, share experiences, and learn about what has and hasn't worked for everyone.

Regarding logging vs tracing, FB's experiences are probably closest to what @beberlei describes, with some of what @bhs notes about sampling and correlation. I'm still building up my mental map of terminology to translate concepts back and forth in my head, so apologies if I misunderstand anything. Also, this is an area we're still currently exploring internally, so things may change etc.

1. In some domains, we pull in point-logged data logged via general logging APIs into the trace, and we're currently looking to expand our use of it given its usefulness where we already do it. Here, we usually don't have all logging go through the tracing API, but rather the trace construction can ingest events and info from a variety of sources (so users use the logging APIs as normal, and annotated points will be added to active traces if they exist, as well as being routed towards whatever else is consuming the log events). This sounds roughly equivalent to the statement that you need the current span in order to attach the log entries if you're using existing logging APIs.
2. Making sure log entries refer to the proper scope of the trace (e.g. request) is important. Right now, the common logging interfaces we try to incorporate mostly exist in heavily scoped environments (web servers / mobile apps) where log calls are by default scoped to the thing we're tracing and it's actually nontrivial to get any log information outside of it, let alone an uncorrelated mixture of log entries and trace structure.
3. It helps that these log calls go through a heavily structured interface already so we don't need to do any parsing or analysis of the log lines in general. We get a timestamp + structured metadata, which corresponds very well with our trace events (I think, though I am not sure, this maps to what you refer to as KV logging as opposed to string logging).
4. The usefulness of the data is in the built-in correlations it provides — even within a single system, the logging may be sampled due to volume, but sampled at the granularity of individual events, which makes drawing conclusions from multi-event correlations really difficult. Having a single timeline consisting of all of these trace + user log events has been helpful in e.g finding traces with a particular set of logged events for further investigation, or looking at relations of events w/in the context of a specific trace. That said, it definitely is not a replacement for logging, as the overall volume of logged events is expected to be larger than the volume of traces, and the sampling issues cut both ways (log events may not want to be sampled at whatever rate we end up tracing, or may have sampling biases due to how we sample for traces)

**tl;dr:** We have found incorporating other logging inside traces useful. It works for us because the places that we do it or are considering it already solve a bunch of the annoying problems with doing it in general. Logging != tracing, but traces are super useful for multi-event correlations over log events

adriancole:
@jmkaldor thx for the overview. interesting approach!

bhs:
I'd like to either close this or modify+merge it "soon" – obviously a (long-)dormant PR, and that's not good.

Much of the actual PR content seems like a helpful clarification, and that's good. However, I think we should remove any suggestion that conventional process-level logging be redirected to OT `Span` "micrologging".

If I had to make a concrete proposal, we would remove the bit mentioned above ^^^, raise the rest of this topic at a cross-language WG meeting, and merge the parts that are not controversial. We can leave the rest out.

yurishkuro:
I am going to merge this long outstanding PR. Despite some (even strong) disagreements on this thread, the actual PR text simply clarifies what is already available in the OT APIs and the experience over the last 2yrs showed that developers do like the micro-logging facility in the API (which is why the OpenTelemetry also supports it).

PR #53, opened 2017-03-13
Document a peer.address tag

### LINKED ISSUE #16: Introduce peer.address standard tag
Continuing the discussion from https://github.com/opentracing/opentracing.io/issues/116

cc: @yurishkuro @bensigelman @dkuebric @cwe1ss

bhs:
Closes #16 

My only question here is whether we want to replace/deprecate the `peer.hostname` and friends. Obviously I voted "no" with this PR, but it's worth thinking about.


yurishkuro:
I think at this point deprecating other peer tags might do more harm than good - there are existing instrumentations that are using them already, and the tracing systems may already be dealing with them. Also those tags might be more efficient if instrumentation knows the concrete types, rather than forcing them back to string. For example, zipkin model has dedicated fields for ipv4, ipv6 & port (and pretty much doesn't support anything else for "peer").

wu-sheng:
@yurishkuro Agree, it is better not remove right now. My only concern is people didn't know the background story as we did, they will be confuse, why setting a network address has two ways.

whiskeysierra:
The current specification only states 

> This may be a "ip:port", a bare "hostname"

After reading the original discussions I believe "hostname:port" is also valid. Is that correct? If yes, what about rules when to omit port and when not?

PR #102, opened 2018-02-23
Add ISSUE_TEMPLATE to specification (copy of opentracing-php repo)

dimtruck:
Adds ISSUE_TEMPLATE to specification repo.  This is an (almost) exact copy of the ISSUE_TEMPLATE in [opentracing-php repo](https://github.com/opentracing/opentracing-php/blob/master/.github/ISSUE_TEMPLATE.md)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,9 +1,9 @@
 <!--
 Welcome to the OpenTracing Specification repo
 
-- Please search for existing issues in order to ensure we don't have duplicate bugs/feature requests.)
-- Please be respectful and considerate of others when commenting on issues)
-- Please provide as much information as possible so we all understand the issue.)
+- Please search for existing issues in order to ensure we don't have duplicate bugs/feature requests.
+- Please be respectful and considerate of others when commenting on issues.
+- Please provide as much information as possible so we all understand the issue.
 - Please don't ask questions here. If you have any question head to our gitter chat https://gitter.im/opentracing/public
 -->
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+
+## Background
+Something that gives context about why this is an issue.
+
+## Problem
+Describe the problem. If the issue is about an improvement you can skip this. If possible, include a description of the impact of the problem.
+
+## Proposal
+A proposal that from your POV would solve the problem or improve the existing situation. It should also include the impact.
```
ndrwrbgs: I'd perhaps expand "POV" as we're in an international community, so abbreviations are sometimes hard

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+<!--
+Welcome to the OpenTracing Specification repo
+
+- Please search for existing issues in order to ensure we don't have duplicate bugs/feature requests.
+- Please be respectful and considerate of others when commenting on issues.
+- Please provide as much information as possible so we all understand the issue.
+- Please don't ask questions here. If you have any question head to our gitter chat https://gitter.im/opentracing/public
```
wu-sheng: As a GitHub hosted project, is "don't allow ask question through Issue" a good idea? I prefer we don't make this rule.

dimtruck: That's fair.  I'll reword it to instead point people to try gitter first before creating an issue for a  question just so that issues would contain more of a specification-centric discussion and not necessarily the `how-to`'s of OpenTracing.

dimtruck: Thinking more about it, I'll remove this line entirely.  Thanks!

yurishkuro: I rather liked the idea to direct people to gitter for how-to questions. But I don't think it should block this pr.

PR #103, opened 2018-02-26
Correct indentation level of an example span

christophetd:
Unless I'm mistaken, `Span B` and `Span C` should be at the same level here, since they are both children of `Span A`.


specification.md
```diff
@@ -50,7 +50,7 @@ Temporal relationships between Spans in a single Trace
  [Span A···················································]
    [Span B··············································]
       [Span D··········································]
-    [Span C········································]
+   [Span C········································]
```
wu-sheng: Span C and B are childOf A, yes. But this graph is about time line, so B and C maybe or maybe not at the same level, which means happens in the same time.

Maybe put the graph in this way will help your understanding:

~~~
Temporal relationships between Spans in a single Trace


––|–––––––|–––––––|–––––––|–––––––|–––––––|–––––––|–––––––|–> time

 [Span A·································································································]
   [Span B···································]
      [Span D·······························]
                                            [Span C···········································]
                                                [Span E·······]        [Span F··] [Span G··] [Span H··]
~~~

christophetd: Makes more sense indeed, thanks!

wu-sheng: @christophetd Do the update by https://github.com/opentracing/specification/pull/105

PR #105, opened 2018-02-27
Update temporal graph in specification.md

wu-sheng:
Make `Temporal relationships between Spans in a single Trace` more sense.

Based on https://github.com/opentracing/specification/pull/103#discussion_r170787736

cc @christophetd


pavolloffay (approved):

yurishkuro:
I actually like the original better. It would be even better if there were causality indications, like an edge from A to both B and C. There's nothing wrong with both of them being the children of A and both starting at different times. This PR changes the graph to a more special case of sequential spans.

wu-sheng:
> This PR changes the graph to a more special case of sequential spans

The graph always visualized spans. I agree new graph is sequential spans case, but I think it is more common, but no so special. In my perspective, most language application runs in sequential way. The old graph shows more about a async process graph, which may confuse people, like #103 

I agree either way is just a case, this is not a right/wrong discussion. I just want to make people more easier to understand when they read the document at first time. @yurishkuro

tedsuo:
The real answer is to use an actual image of a graph, with edges and a legend. I don't really feel like this is better or worse.

jpkrohling:
> The real answer is to use an actual image of a graph

This.

PR #106, opened 2018-02-28
Versioning process for the OpenTracing specification

tedsuo:
## RFC Proposal
https://github.com/opentracing/specification/blob/tedsuo/rfc-process/rfc_process.md
Defines a more formal versioning process for the OpenTracing specification.

## RFC Template
https://github.com/opentracing/specification/blob/tedsuo/rfc-process/rfc_template.md
To be used for starting proposals.

## Example RFC
https://github.com/opentracing/specification/pull/107 [Trace-Parent Accessors](https://github.com/opentracing/specification/pull/107)
A real change request, done in the format defined within this pull request.

## Inspiration and prior art
* Couchbase: https://github.com/couchbaselabs/sdk-rfcs/
* IETF
* W3C

### DIFF
--- a/rfc_process.md
+++ b/rfc_process.md
@@ -65,7 +65,7 @@ Consensus is used to determine that all of the abstract proposal work has been c
 
 ## Test
 * RFC status changed from `Draft` to `Test`.
-* Release Candidate per major language is created.
+* Release Candidate per supported language is created.
 * API conventions are coordinated between languages.
 * New risks and ambiguities may lead to proposal changes.
 
--- a/rfc_process.md
+++ b/rfc_process.md
@@ -3,13 +3,13 @@ The OpenTracing specification consists of a cross-language, abstract interface, 
 
 In order effect changes to the OpenTracing specification, all interfaces are developed together in a three step process:
 
-* DRAFT: An RFC is drafted to define and justify the desired changes.
-* TEST: A new version of each language interface is designed, tested, and released.
-* ACCEPTED: The abstract interface is versioned to reflect the final changes.
+* **DRAFT:** An RFC is drafted to define and justify the desired changes.
+* **TEST:** A new version of each language interface is designed, tested, and released.
+* **ACCEPTED:** The abstract interface is versioned to reflect the final changes.
 
 # Proposal Components
 
-## RFC (Request for Change)
+## RFC (Request for Comments)
 Proposals begin as a single document, written in markdown and committed to the `/rfc` directory.
 
 An RFC consists of the following:
@@ -47,7 +47,7 @@ Each language interface implements the abstract interface contained in the RFC.
 A complete release candidate consists of the following:
 * Repository branch containing the new version.
 * Packages for installing the new version.
-* Tracers which bind to the new version.
+* More than one tracer which binds to the new version.
 * Major instrumentation ported to the new version.
 
 # Proposal Lifecycle
@@ -69,7 +69,9 @@ Consensus is used to determine that all of the abstract proposal work has been c
 * API conventions are coordinated between languages.
 * New risks and ambiguities may lead to proposal changes.
 
-Once the draft proposal is finalized, language implementation begins. Language maintainers begin creating a release candidate on a branch of the language repository. Links to the release candidates, along with their status, are recorded in the tracking issue.
+Once the draft proposal is finalized, language implementation begins. Language maintainers begin creating a release candidate on a branch of the language repository. Links to the release candidates, along with their status, are recorded in the tracking issue. Multiple tracers create a branch which is compatible with new API, along with important instrumentation libraries. These assest are then used to vet the new API, exploring edge cases and potential issues.
+
+At this stage, issues and further refinements are surfaced and debated via tested examples, as English often is too imprecise to address the finer points of API design.
 
 ## Accepted
 * Remaining language interfaces are released.
--- a/rfc_process.md
+++ b/rfc_process.md
@@ -1,0 +1,83 @@
+# Versioning Process for the OpenTracing Specification
+The OpenTracing specification consists of a cross-language, abstract interface, along with an attendant set of language-specific interfaces. Every interface is versioned, with each language interface supporting a version of the abstract interface.
+
+In order effect changes to the OpenTracing specification, all interfaces are developed together in a three step process:
+
+* DRAFT: An RFC is drafted to define and justify the desired changes.
+* TEST: A new version of each langauage interface is designed, tested, and released.
+* ACCEPTED: The abstract interface is versioned to reflect the final changes.
+
+# Proposal Components
+
+## RFC (Request for Change)
+Proposals begin as a single document, written in markdown and committed to the `/rfc` directory.
+
+An RFC consists of the following:
+
+* Status, date, and author.
+* Problem statement.
+* Historical background.
+* Abstract interface.
+* Use cases.
+* Risk Assesment.
+
+An RFC has the following qualities:
+
+* The issue is clearly defined, and within the scope of the OpenTracing charter.
+* The utility to the OT community is illustrated with concrete use cases. It should be clear not only what will change, but how the new interface is expected to be used.
+* Changes to the abstract interface are clearly defined. Language-specific interfaces will be implemented in a later phase,but the abstract interface must be clear enough that the changes could be ported to many languages without much deviation.
+* All expected behavior should be documented, in sufficient detail such that language-specific tests and examples can be generated from the descriptions.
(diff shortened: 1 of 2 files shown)


rfc_process.md
```diff
@@ -0,0 +1,83 @@
+# Versioning Process for the OpenTracing Specification
+The OpenTracing specification consists of a cross-language, abstract interface, along with an attendant set of language-specific interfaces. Every interface is versioned, with each language interface supporting a version of the abstract interface.
+
+In order effect changes to the OpenTracing specification, all interfaces are developed together in a three step process:
+
+* DRAFT: An RFC is drafted to define and justify the desired changes.
+* TEST: A new version of each langauage interface is designed, tested, and released.
```
yurishkuro: typo: langauage

rfc_process.md
```diff
@@ -0,0 +1,83 @@
...
+* Links to relevant artifacts, such as release candidates.
+* Links to meetings and other discussion channels.
+
+## Release Candidates
+Each language interface implements the abstract interface contained in the RFC.
+
+A complete release candidate consists of the following:
+* Repository branch containing the new version.
+* Packages for installing the new version.
+* Tracers which bind to the new version.
```
cwe1ss: How should we define/choose which tracer should try to implement a feature? Do we just rely on the fact that people here are working on tracers anyway and will implement features in "their" tracers? Also, tracers are more or less separate projects with separate release cyclesoutside of our direct control (but tied to us of course). Maybe it would be easier if every language had the "basic tracer" which could be used to show the usage of the feature?

rfc_process.md
```diff
@@ -0,0 +1,83 @@
...
+Consensus is used to determine that all of the abstract proposal work has been completed, and the investigation should move to language implementations.
+
+## Test
+* RFC status changed from `Draft` to `Test`.
+* Release Candidate per major language is created.
+* API conventions are coordinated between languages.
+* New risks and ambiguities may lead to proposal changes.
+
+Once the draft proposal is finalized, language implementation begins. Language maintainers begin creating a release candidate on a branch of the language repository. Links to the release candidates, along with their status, are recorded in the tracking issue.
+
```
cwe1ss: This no longer explains the previously mentioned point `Tracers which bind to the new version.`. See my previous comment for questions regarding it.

rfc_process.md
```diff
@@ -0,0 +1,83 @@
...
+## Accepted
+* Remaining language interfaces are released.
+* Abstract interface is versioned and updated.
+* Documentation is updated.
+* RFC status changed from `Test` to `Accepted`.
+* Tracking issue for the proposal is closed.
+
+After a quorum of language interfaces have been tested and released, consensus may be reached that a proposal has been finalized. At this point, the abstract interface contained in the `specification.md` document is updated to reflect the final changes, based on the language in the RFC docment and release candidates.
```
cwe1ss: Isn't this already done via the "Accepted" stage that says that the "Abstract interface is versioned and updated"?

rfc_process.md
```diff
@@ -0,0 +1,83 @@
...
+
+In addition to the specification, the OpenTracing website and other major docmentation efforts are updated at this time to reflect the latest version. Once all work is complete, the tracking issue is closed.
```
cwe1ss: typo: docmentation

rfc_template.md
```diff
@@ -0,0 +1,18 @@
+# 0001: RFC_NAME
```
cwe1ss: Do we really need numbered RFCs? They will have separate lifetimes and later RFCs might be released earlier etc. Also, it might be hard to know what the next free number is. (One might have to look through open and closed issues etc.)

As a different solution, e.g. in the "C# language" repository, they just [use names for their proposals](https://github.com/dotnet/csharplang/tree/master/proposals).

rfc_template.md
```diff
@@ -0,0 +1,18 @@
...
+
+**Updated:** January 01, 2018
```
cwe1ss: Since files are committed markdown files with a git history, it might not be necessary to use a "Updated" column. Maybe it's enough to keep dates for state changes?

rfc_template.md
```diff
@@ -0,0 +1,18 @@
...
+
+# Specification Changes
+Changes to the abstract interface must be clearly defined. Language-specific interfaces will be implemented in a later phase, and the abstract interface contained here must be clear enough such that the changes can be ported to many languages without much deviation.
+
+All expected behavior should be documented, in sufficient detail such that language-specific tests and examples can be generated from the descriptions.
+
+# Use Cases
+Concrete use cases must be provided, in order to guide testing and interface design, and assist reviewers.
+
+# Risk Assesment
```
cwe1ss: typo: I think it should be "Assessment"

rfc_process.md
tedsuo: In general, I believe we should use "rough consensus" for making these decisions. No one can use an API without at least one tracer and one complete set of instrumentation (framework plus RPC library). So an API can't really be examined without these. Ideally, a new release candidate is not made official without at least support from some tracers and instrumentation, or it is useless to people. As to which ones and how many, that should be a case by case basis. For example, which tracers support C# might be different from those which support C++.

We definitely need to revisit the Basic Tracers and make them more useful, but I don't think they are a substitute.

rfc_template.md
tedsuo: Should be clear what the next free number is when you commit (may be different than when you started, of course). Hopefully we are not accepting proposals so fast that there is some kind of race condition. :)

That said, I am not wedded to numbered RFCs. For me, it's simply a matter of how we would like the files organized, so they aren't just in one big arbitrary pile. Possibly directories like C# uses would also be a good approach.

rfc_template.md
tedsuo: Yeah, could be fine! I'm down to remove this.

rfc_process.md
tedsuo: Not sure what distinction you are making. This is a description of the Accepted stage.

rfc_process.md
tedsuo: Ok, sounds like more detail is needed to describe the Test stage.

rfc_process.md
cwe1ss: my bad - I thought this was a new paragraph.

tedsuo:
I'd like to merge this in on Friday, March 16th, after the Cross-Language working group meeting. If you have objections or questions, please give them this week.

rfc_process.md
```diff
@@ -0,0 +1,85 @@
...
+
+In order effect changes to the OpenTracing specification, all interfaces are developed together in a three step process:
+
+* DRAFT: An RFC is drafted to define and justify the desired changes.
+* TEST: A new version of each language interface is designed, tested, and released.
+* ACCEPTED: The abstract interface is versioned to reflect the final changes.
+
+# Proposal Components
+
+## RFC (Request for Change)
```
jpkrohling: Usually, "RFC" is a "request for comments", not change. Did you really mean "change"?

rfc_process.md
```diff
@@ -0,0 +1,85 @@
...
+* Abstract interface.
+* Use cases.
+* Risk Assessment.
+
+An RFC has the following qualities:
+
+* The issue is clearly defined, and within the scope of the OpenTracing charter.
+* The utility to the OT community is illustrated with concrete use cases. It should be clear not only what will change, but how the new interface is expected to be used.
+* Changes to the abstract interface are clearly defined. Language-specific interfaces will be implemented in a later phase, but the abstract interface must be clear enough that the changes could be ported to many languages without much deviation.
+* All expected behavior should be documented, in sufficient detail such that language-specific tests and examples can be generated from the descriptions.
```
jpkrohling: +1

rfc_process.md
```diff
@@ -0,0 +1,85 @@
...
+
+A draft proposal is first submitted as a pull request against the specification repository. 
+
+Once there is consensus that the pull request meets the minimum criteria for the beginnings of a draft, and has strong potential to be accepted, the pull request is merged. A tracking issue is created at this point to manage the proposal. The draft RFC is then refined via more pull requests.
+
+Consensus is used to determine that all of the abstract proposal work has been completed, and the investigation should move to language implementations.
+
+## Test
+* RFC status changed from `Draft` to `Test`.
+* Release Candidate per major language is created.
```
jpkrohling: Do we have a list of "major languages" defined somewhere?

tedsuo: good point, "supported language" is a better term.

tedsuo:
@cwe1ss @jpkrohling thank you for your feedback. I've made changes based on your suggestions, and I'm going to merge. Feel free to make additional PRs if you find further changes.

PR #108, opened 2018-03-07
Initial draft of the ScopeManager proposal.

carlosalberto:
# Scope Manager

**Current State:** Draft  
**Author:** [carlosalberto](https://github.com/carlosalberto)

In the OpenTracing specification, under the "Optional API Elements" section, it is mentioned languages may choose to provide utilities to pass an active **Span** around a single process.

Upon many iterations and feedback from several contributors, the Java 0.31 API defined the new **Scope Manager** concept, which is a simple and explicit way to manage the active **Span** at a given call-context point. This document intends to standardize this concept, so other supported languages and platforms can leverage it, with their respective semantical differences.

# Technical background

For any thread, at most one **Span** may be "active". Of course, there may be many other **Spans** involved with the thread which are (a) started, (b) not finished, and yet (c) not "active": perhaps they are waiting for I/O, blocked on a child **Span**, or otherwise off of the critical path.

For platforms where the call-context is explicitly propagated down the execution chain -such as `Go`-, such context can be used to store the active **Span** at all times.

For platforms not propagating the call-context, it's inconvenient to pass the active **Span** from function to function manually, so OpenTracing should require, for those platforms, that **Tracer** contains a **Scope Manager** that grants access to the active **Span** through a container, called **Scope** (using some call-context storage, such as thread-local or coroutine-local).

# Specification Changes

New `ScopeManager` and `Scope` interfaces are added to the specification, and the `Tracer` interface is extended to support creation of `Span`s that are automatically set as the **active** one for the current context.

## ScopeManager

The `ScopeManager` interface allows setting the active `Span` in a call-context storage section, and has the following members:

* **activate**, capability to set the specified `Span` as the active one for the current call-context, returning a `Scope` containing it. A required boolean parameter **finish span on close** will mark whether the returned `Scope` should, upon deactivation, **finish** the contained `Span`.
* **active**, the `Scope` containing the current active `Span` if any, or else null/nothing.

## Scope

The `Scope` interface acts as a container of the active `Span` for the current-call context, and has the following members:

* **span**, the contained active `Span` for this call-context. It will never be null/nothing.
* **close**, marking the end of the active period for the current `Span`, and optionally **finishing** it. Calling it more than once leads to undefined behavior.

If the language supports some kind of auto finishing statement (such as `try` for Java, or `with` for Python), `Scope` should adhere to such convention. Additionally, `Scope` is not guaranteed to be thread-safe.

## Tracer changes

The `Tracer` interface will be extended with:

* **scope manager**, the `ScopeManager` tracking the active `Span` for this instance.
* **start active span**, a new behavior for starting a new `Span`, which will be automatically marked as active for the current call-context. It will return a `Scope`. A parameter **finish span on close** will mark whether the `Scope` should, upon deactivation, **finish** the contained `Span`. A default value for this parameter may be provided, depending on the suitability for the language and its use cases.
* Both **start span** and **start active span** will implicitly use any active `Span` as the parent for newly created `Span`s (with a `ChildOf` relationship), unless the parent is explicitly specified, or (the new) **ignore active span** parameter is specified (in which case the resulting `Span` will have no parent at all).

# Use Cases

## Single-threaded operations.

The active `Span` will be accessible at all times through the `Tracer`, without any need to pass the `Span` around when creating children.

## Multi-threaded operations.

The active `Span` will be accessible at all times, probably using thread-local storage, and it will be possible to pass `Span` instances between threads and manually manage its active period, and have full, manual control on when it should be finished.

# Risk Assessment

The following risks have been identified:

* This change will mean API breakage, as both **start span** and **start active span** behaviors of `Tracer` will implicitly use any active `Span` as the parent for newly created `Span`s (previously, if no parent was specified, all created `Span`s were parentless).
* For platforms having a prior version with a different in-context propagation (as it happened to the 0.30 Java version, which included a concept called `ActiveSpanSource`, which used reference-count to handle the lifetime of `Span`s), either a full migration will be needed, or at least a shim/bridge layer should be provided to ease with the proposed changes.
* There can be languages and platforms for which, given their specific threading or memory models, implementing this changes will either be not sufficient or redundant.


carlosalberto:
CC @palazzem @cwe1ss as contributors who have been working on `ScopeManager` implementations for other languages ;)

rfc/0002_scope_manager.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Specification Changes
+
+New `ScopeManager` and `Scope` interfaces are added to the specification, and the `Tracer` interface is extended to support creation of `Span`s that are automatically set as the **active** one for the current context.
+
+## ScopeManager
+
+The `ScopeManager` interface allows setting the active `Span` in a call-context storage section, and has the following members:
+
+* **activate**, capability to set the specified `Span` as the active one for the current call-context, returning a `Scope` containing it. A parameter (still to decide whether it will be optional or required) will mark whether the returned `Scope` should, upon deactivation, **finish** the contained `Span`.
```
cwe1ss: My argument for keeping the parameter required here is that the `activate` method gets an **existing** span so it should also be up to the caller whether or not the span should be finished when the scope is closed.

rfc/0002_scope_manager.md
```diff
@@ -0,0 +1,61 @@
...
+* **active**, the `Scope` containing the current active `Span` if any, or else null/nothing.
+
+## Scope
+
+The `Scope` interface acts as a container of the active `Span` for the current-call context, and has the following members:
+
+* **span**, the contained active `Span` for this call-context. It will never be null/nothing.
+* **close**, marking the end of the active period for the current `Span`, and optionally **finishing** it. Calling it more tan once leads to undefined behavior.
```
cwe1ss: nit: "than" instead of "tan".

rfc/0002_scope_manager.md
```diff
@@ -0,0 +1,61 @@
...
+
+If the language supports some kind of auto finishing statement (such as `try` for Java, or `with` for Python), `Scope` should adhere to such convention. Additionally, `Scope` is not guaranteed to be thread-safe.
+
+## Tracer changes
+
+The `Tracer` interface will be extended with:
+
+* **scope manager**, the `ScopeManager` tracking the active `Span` for this instance.
+* **start active span**, a new behavior for starting a new `Span`, which will be automatically marked as active for the current call-context. It will return a `Scope`.
```
cwe1ss: This doesn't mention the `finishSpanOnClose` parameter. In C#, we'd like to make the default value `finishSpanOnClose: true` because this is the main use case there. 

My argument for adding a default behavior of `finishSpanOnClose: true` on this method on all languages is that this method creates the scope **and** the span so it is "responsible" for the span. Closing the scope should therefore also finish the scope (unless defined differently by the developer).

rfc/0002_scope_manager.md
```diff
@@ -0,0 +1,61 @@
...
+* Both **start span** and **start active span** will implicitly use any active `Span` as the parent for newly created `Span`s (with a `ChildOf` relationship), unless the parent is explicitly specified, or (the new) **ignore parent span** parameter is specified (in which case the resulting `Span` will have no parent at all).
```
cwe1ss: "ignore **active** span" instead of "ignore **parent** span". (The method name in Java/C# is `ignoreActiveSpan`).

rfc/0002_scope_manager.md
carlosalberto: Definitely, not as 'obvious' as compared to **start active span**. I will leave it as it is for now.

rfc/0002_scope_manager.md
carlosalberto: Agreed - although the mention of such parameter will be done on a more high-level way ;)

carlosalberto:
Updated the RFC to reflect (as per @cwe1ss feedback), **finish span on close** mention for both **start active span** and **ScopeManager's activate**.

rfc/scope_manager.md
```diff
@@ -0,0 +1,61 @@
...
+
+In the OpenTracing specification, under the "Optional API Elements" section, it is mentioned languages may choose to provide utilities to pass an active **Span** around a single process.
+
+Upon many iterations and feedback from several contributors, the Java 0.31 API defined the new **Scope Manager** concept, which is a simple and explicit way to manage the active **Span** at a given call-context point. This document intends to standardize this concept, so other supported languages and platforms can leverage it, with their respective semantical differences.
+
+# Technical background
+
+For any thread, at most one **Span** may be "active". Of course there may be many other **Spans** involved with the thread which are (a) started, (b) not finished, and yet (c) not "active": perhaps they are waiting for I/O, blocked on a child **Span**, or otherwise off of the critical path.
+ 
+It's inconvenient to pass an active **Span** from function to function manually, so OpenTracing should require that every **Tracer** contains a **Scope Manager** that grants access to the active **Span** through a container, called **Scope**.
```
jpkrohling: I'd rewrite this as something like: "For applications where the whole unit of work happens in a single thread, it's more convenient to store the active span in a thread-local variable than to pass the span context to all involved functions."

I wouldn't say that passing the context down the chain is "inconvenient" (it's why Go's context was made for, IIRC). Rather, it's *more* convenient to not pass this context around, whenever possible. Unfortunately, new generation applications using reactive/stream patterns can't make use of thread-local vars, so, passing down the context is the right approach.

carlosalberto:
@jpkrohling Hey, updated the draft to reflect the `Go` case - decided to leave (for now) the thread and the reactive parts out as they could be more of implementation detail (i.e. Java works 'just fine' using the `ScopeManager` concept, but _some_ of their frameworks may find this insufficient, etc).

Hope that makes things clear for the remaining of the process, anyway ;)

jpkrohling:
I'm not sure we can say that Go has "call-context propagated down the execution chain". What I meant is that it's common in Go to pass down a context object[1], and such context object would contain a span context[2]. If that's indeed what you mean, I'm +1 to the changes.

1 - https://golang.org/pkg/context/
2 - https://github.com/opentracing/opentracing-go#creating-a-span-given-an-existing-go-contextcontext

carlosalberto:
Hey @jpkrohling thanks again for the feedback ;) 

So yes, I had briefly consulted the documentation for `Go` and how they use `Context` to store the active `Span` - so my question is more about how you think I should write this part, i.e. not `call-context propagated down the execution chain` (to me, as a non-Go developer, it **looks** like the context is propagated down the chain, even if explicitly).

Let me know - the clearer we can define this part, the better ;)

jpkrohling:
I think something similar to my initial proposal would describe the scenario in a way that is independent of the language:

> For applications where the whole unit of work happens in a single thread, it's more convenient to store the active span in a thread-local variable than to pass the span context to all involved functions

Otherwise, your version would also work. Perhaps it would make sense to just add the word "explicitly" here:

> For platforms where the call-context is explicitly propagated down the execution chain...

billowqiu:
any implement suggest in cpp?

PR #113, opened 2018-03-29
Rephrase 'xor'. Add missing parenthesis

pgadige:
'xor' is corrected to 'or'
adds missing closing parenthesis

Signed-off-by: Pooja Gadige <poojagadige@gmail.com>

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -171,7 +171,7 @@ Required parameters
 Both injection and extraction rely on an extensible **format** parameter that dictates the type of the associated "carrier" as well as how a `SpanContext` is encoded in that carrier. All of the following **format**s must be supported by all Tracer implementations.
 
 - **Text Map**: an arbitrary string-to-string map with an unrestricted character set for both keys and values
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -238,7 +238,7 @@ Required parameters
 
 - The **baggage key**, a string
 
-**Returns** the corresponding **baggage value**, xor some indication that such a value was missing.
+**Returns** the corresponding **baggage value**, or some indication that such a value was missing.
```
jpkrohling: `xor` might have been intentional here, indicating an ["exclusive or"](https://en.wikipedia.org/wiki/Exclusive_or). I don't think your changes cause any loss of meaning here, but an alternative might have been:

> **Returns** either the corresponding **baggage value**, or some indication that such a value was missing.

pgadige: Thank you for the valuable feedback @jpkrohling 
I shall revert it to 'xor', since it's more specific and unambiguous as compared to using 'either....or...'

jpkrohling: I personally prefer "either...or", as it gives the right meaning. I personally tend to dislike "xor" on documentation.

pgadige: Yay! I do too! :+1: Will change it right away.

PR #118, opened 2018-05-08
Remove dash in `Trace-Parent`

inlined:
Both the [repo](https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md) and [W3C Spec](https://w3c.github.io/distributed-tracing/report-trace-context.html) for Distributed Tracing describe the `traceparent` header without dashes.


tylerbenson:
I would suggest following their lowercase format too.

inlined:
Done. If you'd like I can also document the only known valid bit field for trace options.

tylerbenson:
I don't have merge rights, but I think that would be a reasonable thing to do.

inlined:
Looks like the typo was fixed in my epic absence.

PR #122, opened 2018-07-05
Set the scope_manager RFC to Test.

carlosalberto:
Initial take on the `ScopeManager` RFC.

Notes:

* Used 'thread or execution unit' in order to abstract things such as coroutines.
* Mentioned `Scope`s are not thread-safe.
* Mentioned the `null` the value (it hadn't been mentioned before in the Specification).
* Light take on the recent `start_active_span()` vs `start_active_scope()` debate.
* Used `ScopeManager.active()` (as opposed to `ScopeManager.activeSpan()`). Prone to change based on further discussion.

Please provide feedback ;)

@tedsuo @palazzem @pavolloffay @pglombardo @yurishkuro @jpkrohling @indrekj @mwear @felixbarny @cwe1ss @adriancole

### DIFF
--- a/rfc/scope_manager.md
+++ b/rfc/scope_manager.md
@@ -1,6 +1,6 @@
 # Scope Manager
 
-**Current State:** Draft  
+**Current State:** Test
 **Author:** [carlosalberto](https://github.com/carlosalberto)
 
 In the OpenTracing specification, under the "Optional API Elements" section, it is mentioned languages may choose to provide utilities to pass an active **Span** around a single process.
--- a/rfc/scope_manager.md
+++ b/rfc/scope_manager.md
@@ -23,15 +23,17 @@ New `ScopeManager` and `Scope` interfaces are added to the specification, and th
 
 The `ScopeManager` interface allows setting the active `Span` in a call-context storage section, and has the following members:
 
-* **activate**, capability to set the specified `Span` as the active one for the current call-context, returning a `Scope` containing it. A required boolean parameter **finish span on close** will mark whether the returned `Scope` should, upon deactivation, **finish** the contained `Span`.
-* **active**, the `Scope` containing the current active `Span` if any, or else null/nothing.
+* **activate**, capability to set the specified `Span` as the active one for the current call-context, returning a `Scope` to later end its active period.
+  An **optional** feature to optionally **finish** the `Span` upon `Scope` deactivation may be provided, depending on its suitability for the language. It would require a boolean parameter **finish span on close**.
+* **active Span**, the `Span` that is currently active, if any, or else null/nothing.
+* **active**, the corresponding `Scope` for the current active `Span` if any, or else null/nothing.
 
 ## Scope
 
 The `Scope` interface acts as a container of the active `Span` for the current-call context, and has the following members:
 
-* **span**, the contained active `Span` for this call-context. It will never be null/nothing.
 * **close**, marking the end of the active period for the current `Span`, and optionally **finishing** it. Calling it more than once leads to undefined behavior.
+* **span** (optional), the related active `Span` for this call-context, depending on its suitability for the language. It will never be null/nothing.
 
 If the language supports some kind of auto finishing statement (such as `try` for Java, or `with` for Python), `Scope` should adhere to such convention. Additionally, `Scope` is not guaranteed to be thread-safe.
 
@@ -40,7 +42,8 @@ If the language supports some kind of auto finishing statement (such as `try` fo
 The `Tracer` interface will be extended with:
 
 * **scope manager**, the `ScopeManager` tracking the active `Span` for this instance.
-* **start active span**, a new behavior for starting a new `Span`, which will be automatically marked as active for the current call-context. It will return a `Scope`. A parameter **finish span on close** will mark whether the `Scope` should, upon deactivation, **finish** the contained `Span`. A default value for this parameter may be provided, depending on the suitability for the language and its use cases.
+* **start active span** (optional), a new behavior for starting a new `Span`, which will be automatically marked as active for the current call-context. It will return a `Scope`.
+  An **optional** feature to optionally finish `Span` upon `Scope` deactivation may be provided, depending on its suitability for the language. It would require a boolean parameter **finish span on close**. A default value for this parameter may be provided, depending again on the suitability for the language and its use cases.
 * Both **start span** and **start active span** will implicitly use any active `Span` as the parent for newly created `Span`s (with a `ChildOf` relationship), unless the parent is explicitly specified, or (the new) **ignore active span** parameter is specified (in which case the resulting `Span` will have no parent at all).
 
 # Use Cases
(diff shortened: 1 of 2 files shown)


specification.md
```diff
@@ -250,6 +277,46 @@ In OpenTracing we force `SpanContext` instances to be **immutable** in order to
...
+- **span**, the `Span` instance to be set as active.
+- **finish Span on close**, a boolean value specifying whether the `Span` should be `Finish`ed upon being deactivated through its container `Scope` being `Close`d.
+
+**Returns** a `Scope` instance containing the newly activated `Span`.
+
+#### Retrieve the active `Span`
+
+There should be no parameters.
+
+**Returns** a `Scope` instance containing the active `Span`, or else `null` if there's none for the current thread or execution unit.
```
tylerbenson: This says it returns a `Scope`, but the title implies returning a `Span`.

carlosalberto: Oh yes, think we should tune that in.

tedsuo (approved):
Generally LGTM

specification.md
```diff
@@ -117,16 +117,30 @@ These can all be valid timing diagrams for children that "FollowFrom" a parent.
...
-(serialize) and `Extract` (deserialize) them across process boundaries.
+The `Tracer` interface creates `Span`s, understands how to `Inject`
+(serialize) and `Extract` (deserialize) them across process boundaries, and optionally contains a `ScopeManager` instance to handle the active `Span`.
 Formally, it has the following capabilities:
 
+#### Retrieve the `ScopeManager` (Optional)
+
+There should be no parameters.
+
+**Returns** the used `ScopeManager` instance to set and retrieve the active `Span`. The mentioned active instance is additinaly used by default as the implicit parent for newly created `Span`s, in case no **references** were provided.
```
palazzem: "the active `Span`" -> "the active `Scope`"? or in general I think it should be clear that the `ScopeManager` retrieves a `Scope` that contains the active `Span`. Thoughts?

specification.md
```diff
@@ -144,11 +158,24 @@ For example, here are potential **operation names** for a `Span` that gets hypot
...
 - An optional explicit **start timestamp**; if omitted, the current walltime is used by default
 - Zero or more **tags**
 
 **Returns** a `Span` instance that's already started (but not `Finish`ed)
 
+#### Start a `Scope` with a new `Span` (Optional)
+
+This operation does the same as **Start a new `Span`**, in addition to setting the newly created `Span` as the active instance for the current thread or execution unit through the contained `ScopeManager` instance.
+
+It is **highly** encouraged that the name for this operation includes a `Scope` sufix, in order to make clear the type of the returned instance.
```
palazzem: `Scope` sufix -> `Scope` suffix?

specification.md
```diff
@@ -250,6 +277,46 @@ In OpenTracing we force `SpanContext` instances to be **immutable** in order to
 
 This is modeled in different ways depending on the language, but semantically the caller should be able to efficiently iterate through all baggage items in one pass given a `SpanContext` instance.
 
+### `ScopeManager`
+
+The `ScopeManager` interface sets and retrieves the active `Span`, working along the `Scope` container interface.
```
palazzem: Related to my previous comment. Here I guess the behavior is clear, but I think we may need to change something in the other paragraph.

specification.md
palazzem: Yes I think it should be `Scope`, regardless the name we have in our implementations.

pavolloffay: It would be good to resolve https://github.com/opentracing/opentracing-java/issues/267  before adding this to the spec

specification.md
carlosalberto: Yeah, I think it makes more sense to make it clear at all moments that `Scope` objects are being returned/used.

specification.md
carlosalberto: Oh, method name ;) Will clarify that.

specification.md
carlosalberto: I'd say it's more a related effort - as part of this very PR, we need to discuss what is described in #267 (as well as https://github.com/opentracing/opentracing-ruby/pull/36).

But yes, in practical terms, we need to figure those two issues above before finally merging this PR 😉

specification.md
```diff
@@ -250,6 +277,46 @@ In OpenTracing we force `SpanContext` instances to be **immutable** in order to
...
+
+#### Retrieve the active `Span`
+
+There should be no parameters.
+
+**Returns** a `Scope` instance containing the active `Span`, or else `null` if there's none for the current thread or execution unit.
+
+### `Scope`
+
+The `Scope` interface acts as a simple container for the active `Span`,and it not guaranteed to be thread-safe. It has the following capabilities:
```
indrekj: `and it not guaranteed`

indrekj:
Maybe not directly related to this PR but anyway:

The doc explains the behavior of Scope / ScopeManager, but I feel like it's missing the `why`: Why is scope / scope manager needed? Why wouldn't `Tracer#activate(span) :: span` work instead? I've been experimenting with opentracing spec for elixir and the scope manager part is confusing to me atm. There's likely an answer in some old PR but it's hard to find it.

I've implemented the scope manager part in [ruby zipkin OT](https://github.com/salemove/zipkin-ruby-opentracing/blob/master/lib/zipkin/scope_manager.rb) and we've used it in production for some time now. So far I haven't encountered a use case where a simple `activate(span) :: span` & `deactivate()` without Scope container wouldn't work as well.

carlosalberto:
Hey @indrekj 

> Why is scope / scope manager needed?

We essentially the concept has been to separate the active `Span` handling from the `Tracer`, API-wise. Now, as we have iterated, the idea has been to provide shortcuts for the most common operations - but stuff is still expected to be in the `ScopeManager` instance ;)

carlosalberto:
Hey all!

I *think* I've updated all the requested bits of this PR, and I think it's ready to go. If you have some time over the next days, let me know - else I will be merging it soon (remember it's still an ongoing effort and can be changed/updated later).

rfc/scope_manager.md
```diff
@@ -1,6 +1,6 @@
 # Scope Manager
 
-**Current State:** Draft  
+**Current State:** Test
```
Jamim: Hello @carlosalberto,
I think you should not remove two ending spaces here.
They are [required to split lines](https://daringfireball.net/projects/markdown/syntax#p).
> When you do want to insert a \<br /\> break tag using Markdown, you end a line with two or more spaces, then type return.

[Currently](https://github.com/opentracing/specification/blob/434e6764129e0eb02d27fd096191ca52cc8bb2c9/rfc/scope_manager.md), you have `Current State` and `Author` on the same line.

yurishkuro: Wouldn't a blank line between them do the same? Linters do not like trailing spaces (they even show up as red in git diff).

Alternatively, in this case it's appropriate to us bullet points for the metadata.

Jamim: > Wouldn't a blank line between them do the same?

No, it wouldn't. The result won't be the same.

Jamim: > Linters do not like trailing spaces

I know, but it's an issue of those linters and `git`.
It's not a `Markdown` issue.

PR #124, opened 2018-07-30
Swap Jonathan Kaldor for Michael Bevilacqua-Linn on OTIAB

yurishkuro:
Jonathan has moved to non-tracing project and nominated Michael as Facebook representative to OTIAB.

cc @opentracing/otiab @opentracing/otsc  @mblinn


yurishkuro:
@tedsuo can you approve?

PR #125, opened 2018-08-17
Change method siganture to ToTraceID and ToSpanID

tedsuo:
This update to the Trace Identifiers RFC addresses issues identified in the last round of exploration. It changes the method signatures from `TraceID` and `SpanID` to `ToTraceID` `ToSpanID`. A small additional clarification on future formats is also included.

## Why `ToTraceID` instead of `TraceID`?

The method names `ToTraceID` and `ToSpanID` were chosen over `TraceID` and `SpanID` for two reasons:

1) To avoid conflicts with tracer implementations, which often use `TraceID` and `SpanID` to expose their identifiers in their native format.
2) To communicate to the user that a format conversion may be occuring, and thus there may be some cost to calling these methods.

### DIFF
--- a/rfc/trace_identifiers.md
+++ b/rfc/trace_identifiers.md
@@ -54,7 +54,7 @@ The method names `ToTraceID` and `ToSpanID` were chosen over `TraceID` and `Span
 ## Future Formats
 In some cases, additional formats may be appropriate, if a language supports multiple common transport formats. Two reasons for additional formats were identified:
 
-1) Avoiding double converstions. Rather than converting from a native format to a string value, and then directly into another format, an additional methods could be added to allow the tracer to do the conversion directly.
+1) Avoiding double converstions. Rather than converting from a native format to a string value, and then directly into another format, an additional method could be added to allow the tracer to do the conversion directly.
 2) If tracing systems converge on common in-memory formats for Trace-Context identifiers, accessors may be added for those formats as well.
 
 ## Backwards Compatibility and Optional Support
--- a/rfc/trace_identifiers.md
+++ b/rfc/trace_identifiers.md
@@ -52,10 +52,10 @@ The method names `ToTraceID` and `ToSpanID` were chosen over `TraceID` and `Span
 2) To communicate to the user that a format conversion may be occuring, and thus there may be some cost to calling these methods.
 
 ## Future Formats
-In some cases, additional formats may be appropriate, if a language supports multiple common transport formats. Two reasons for additional formats were identified:
+In some cases, additional identifier formats may be added, besides `string`. Two reasons for additional formats were identified:
 
-1) Avoiding double converstions. Rather than converting from a native format to a string value, and then directly into another format, an additional method could be added to allow the tracer to do the conversion directly.
-2) If tracing systems converge on common in-memory formats for Trace-Context identifiers, accessors may be added for those formats as well.
+1) Avoiding double converstions. Rather than converting from a native format to a string value, and then directly into another format, an additional method could be added to allow the tracer to do the conversion directly. For example, using identifiers as byte arrays may turn out to be popular, and may incur an extra conversion cost from string.
+2) If tracing systems converge on a common trace propagation format, such as [Trace-Context](https://github.com/w3c/trace-context), accessors may be added for that format as well.
 
 ## Backwards Compatibility and Optional Support
 The OpenTracing specification does not currently require trace and span identifiers. To continue support for existing tracers, the empty string value can be returned when no ID has been set.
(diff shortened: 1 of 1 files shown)


yurishkuro:
@tedsuo there were also discussions on some PRs about the range of values returned from those methods, like having them to be a hex string, or being URL-safe. I don't think we had a detailed discussion about that. 

Arguments for hex string - the premise of this whole RFC was the agreement reached at W3C WG between all vendors that they should be able to express their trace/span IDs as opaque, 16 and 8 byte arrays respectively. This was different from the original design of SpanContext when there was no such agreement and there as a concern that some vendors wouldn't be able to expose the IDs (btw maybe worth adding this explanation to the RFC). So given the W3C agreement as a premise, why not follow its format recommendation as well and be more prescriptive about the outputs of ToTraceID/ToSpanID ?

tedsuo:
@yurishkuro I think that is an interesting point. Do existing systems all conform to the new W3C standard for identifiers format?

There are separate design decision around adding direct access to W3C Trace-Context data. It would require at least the following:
- Tracers actually adopting these headers and running in Trace-Context mode.
- A library implementing Trace-Context which all tracers share, in order to avoid silly conversions and type mismatches.
- More thought into versioning. If the Trace-Context spec versions, what happens to the OT interface?

My suggestion would be that we add `ToTraceID` and `ToSpanID` as loosely defined as possible. That could still mean setting a maximum expected length and defining a character set. But if we want to expose the Trace-Context data format specifically, it should be a separate RFC, as the requirements for that are very different.

I'm fine with adding this interface now, and a second one later. But if we want to instead **drop** this RFC in favor of **just** having a TraceContext accessor, we should move very quickly on making a proposal for that. 

In any case, I would suggest keeping this PR tightly scoped to just updating the naming convention, since that's a separate issue which I feel we have resolved. We can make a follow up PRs discussing content restrictions on the returned value, etc.

yurishkuro:
So my concern was not about adopting W3C proposal, but about setting the expectations on the output values. These expectations may take the form of a _recommendation_, rather than a requirement. For example, one of the reasons people want trace IDs is to log them. Logs are often formatted in JSON. So if suddenly the output value of ToTraceID is a string with a double-quote in it (for the sake of argument), then it makes JSON more difficult.

If the output format is only recommendation, consumers still need to take appropriate measures to encode the values properly in whatever medium they are serializing them (e.g. URL-encoding, etc.) But at least a recommendation of a hex string makes the encoding no-op in most cases, and even allows straightforward conversion to binary. Once W3C finalizes the proposal (and it doesn't seem likely that hex format will be overturned there), most OT tracers will be already exposing IDs in the compatible format.

Anyway, I am fine with not prescribing any string format at this stage.

rfc/trace_identifiers.md
```diff
@@ -44,10 +44,18 @@ A string is preferred over other formats for the following reasons:
...
-If tracing systems converge on common in-memory formats for Trace-Context identifiers, accessors may be added for those formats as well.
+The method names `ToTraceID` and `ToSpanID` were chosen over `TraceID` and `SpanID` for two reasons:
+
+1) To avoid conflicts with tracer implementations, which often use `TraceID` and `SpanID` to expose their identifiers in their native format.
+2) To communicate to the user that a format conversion may be occuring, and thus there may be some cost to calling these methods.
+
+## Future Formats
+In some cases, additional formats may be appropriate, if a language supports multiple common transport formats. Two reasons for additional formats were identified:
+
+1) Avoiding double converstions. Rather than converting from a native format to a string value, and then directly into another format, an additional methods could be added to allow the tracer to do the conversion directly.
```
erabug: s/methods/method

rfc/trace_identifiers.md
```diff
@@ -44,10 +44,18 @@ A string is preferred over other formats for the following reasons:
...
-In some cases, additional formats may be appropriate, if a language supports multiple common transport formats. Rather than manually converting the string value to another format, an additional accessors could be added to allow the tracer to do the conversion.
+## Why `ToTraceID` instead of `TraceID`?
 
-If tracing systems converge on common in-memory formats for Trace-Context identifiers, accessors may be added for those formats as well.
+The method names `ToTraceID` and `ToSpanID` were chosen over `TraceID` and `SpanID` for two reasons:
+
+1) To avoid conflicts with tracer implementations, which often use `TraceID` and `SpanID` to expose their identifiers in their native format.
+2) To communicate to the user that a format conversion may be occuring, and thus there may be some cost to calling these methods.
+
+## Future Formats
```
erabug: Changing the header from "alternate" to "future" seems to say "we may add other methods to the official spec in the future for these reasons" rather than just "current tracing implementations may have other methods for these reasons". Was that your intention? If so, we could say so more explicitly.

PR #128, opened 2018-09-12
RFC: Closeable Tracers

austinlparker:
This proposal builds off of discussions primarily in the Java API surrounding the inclusion of a **Close** method in that library. The primary goal of adding this to the specification is to provide a clean decoupling of OpenTracing `Tracer`s from their concrete implementations in a consistent way across platforms and languages.

See https://github.com/opentracing/opentracing-java/issues/250 for further recent discussion.

### DIFF
--- a/rfc/closeable_tracer.md
+++ b/rfc/closeable_tracer.md
@@ -1,0 +1,39 @@
+# Closeable Tracers
+
+**Current State:** Draft  
+**Author:** [austinlparker](https://github.com/austinlparker)
+
+This proposal builds off of discussions primarily in the Java API surrounding the inclusion of a **Close** method in that library. The primary goal of adding this to the specification is to provide a clean decoupling of OpenTracing `Tracer`s from their concrete implementations in a consistent way across platforms and languages.
+
+# Background
+Currently, a `Tracer` has three major responsibilities - creating `Span`s and performing de/serialization of them across boundaries via `Extract` and `Inject`. In practice, a concrete implementation of a `Tracer` has more responsibilities; for example, appending created `Span`s to a buffer, or flushing that buffer through a wire format to some listening system that records traces. One very common method implemented by a concrete `Tracer` is `Close`. The implementation, and utility, of this method varies by language. In general,
+calling `Close` on an instantiated `Tracer` will finalize traces and release any resources held by the `Tracer`. 
+
+Discussions (See [this thread](https://github.com/opentracing/opentracing-java/issues/250) for example) around the inclusion of a `Close` method in Java have crystallized around including this feature in the Java interface. Specifically, there would be value in having `Close` specified on the OpenTracing interface in order to prevent a cast to the concrete tracer, which cannot occur in a situation where the implementation is being loaded dynamically. This situation exists in C# as well, where a similar pattern of try-with-resource (called `using`) exists.
+
+# Specification Changes
+The `Tracer` interface is extended with the following:
+
+* **Close**, a method that is responsible for closing the active tracer. A `Tracer` that has been closed should not be able to record any new spans. For `Tracer`s that record `Span`s for persistance, calling **Close** should flush the current in-memory collection to the configured persistance store. In the event that there is no logical justfication for this method on an implementing `Tracer`, it can be considered a NoOp other than setting the **isClosed** property.
+
+* **isClosed**, a boolean property that indicates that the `Tracer` has been closed. If this is **true** it indicates no further `Span`s should be written to this `Tracer`. 
+
+# Use Cases
+The primary use case of this enhancement would be to reduce vendor-specific code for developers who are adding OpenTracing to their code. In lieu of having to perform potentially unsafe casts of a `Tracer` to access it's implementation-specific `Close` method, the interface would ensure that it was available regardless.
+
+In addition, languages that support try-with-resource or using patterns (Java and C# respectively) will be able to have their `Tracer` implementations implement the respective interfaces in those languages to allow for a more natural usage of the concrete `Tracer`. For example, in C# the following initialization code could be used -
+
+```csharp
+using (var tracer = new Tracer()) {
+    GlobalTracer.Register(tracer);
+    // perform more initialization and setup
+    // start main loop or do work
+}
+// if an exception occurs, the Dispose method of tracer would be called to enable for cleanup and batching/sending of Spans.
+```
+
+# Risk Assessment
+The following risks have been identified:
+
+* While this is an additive change, it could potentially require changes in existing implementations to meet the new interface signature, especially if they already have a **Close** method in their `Tracer`.
+* Since the behavior of **Close** is variable between stateful and stateless tracers, potential for confusion/misuse by end users exists. This is addressable via documentation and education, however.
--- a/rfc/closeable_tracer.md
+++ b/rfc/closeable_tracer.md
@@ -14,7 +14,7 @@ Discussions (See [this thread](https://github.com/opentracing/opentracing-java/i
 # Specification Changes
 The `Tracer` interface is extended with the following:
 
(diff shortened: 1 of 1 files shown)


rfc/closeable_tracer.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Specification Changes
+The `Tracer` interface is extended with the following:
+
+* **Close**, a method that is responsible for closing the active tracer. A `Tracer` that has been closed should not be able to record any new spans. For `Tracer`s that record `Span`s for persistance, calling **Close** should flush the current in-memory collection to the configured persistance store. In the event that there is no logical justfication for this method on an implementing `Tracer`, it can be considered a NoOp other than setting the **isClosed** property.
+
+* **isClosed**, a boolean property that indicates that the `Tracer` has been closed. If this is **true** it indicates no further `Span`s should be written to this `Tracer`.
```
yurishkuro: Why does this need to be in the spec? None of the use cases that were raised around close() mentioned the need for this property as a public API.

rfc/closeable_tracer.md
```diff
@@ -0,0 +1,39 @@
...
+
+Discussions (See [this thread](https://github.com/opentracing/opentracing-java/issues/250) for example) around the inclusion of a `Close` method in Java have crystallized around including this feature in the Java interface. Specifically, there would be value in having `Close` specified on the OpenTracing interface in order to prevent a cast to the concrete tracer, which cannot occur in a situation where the implementation is being loaded dynamically. This situation exists in C# as well, where a similar pattern of try-with-resource (called `using`) exists.
+
+# Specification Changes
+The `Tracer` interface is extended with the following:
+
+* **Close**, a method that is responsible for closing the active tracer. A `Tracer` that has been closed should not be able to record any new spans. For `Tracer`s that record `Span`s for persistance, calling **Close** should flush the current in-memory collection to the configured persistance store. In the event that there is no logical justfication for this method on an implementing `Tracer`, it can be considered a NoOp other than setting the **isClosed** property.
```
yurishkuro: There have been many questions about the behavior of span.finish() when called more then once, which was left as undefined in the spec. I think we can be more specific with the close() method. E.g. that it either must be idempotent, or throw an error.

rfc/closeable_tracer.md
austinlparker: It seemed to me that there needed to be some way to indicate to a consumer that a tracer has been closed. On consideration, it does seem like something that shouldn't be exposed and add a discussion about 'what happens when you try to use a closed Tracer' to the Use Case section?

rfc/closeable_tracer.md
austinlparker: Agreed. I think idempotency makes more sense than increasing the amount of things that can cause errors.

yurishkuro:
What happens to spans started or even finished after the tracer is closed is a good question. We have seen cases where async callbacks were executing and finishing spans even after the tracer was closed, so sometime the application does not have full control over that. Ie we were seeing crashes when the app was just shutting down.

There's a large number of possible states we need to consider.

austinlparker:
The case of 'what happens if I try to start a span after closing a tracer' seems straightforward enough, the tracer should check to see if it's been closed. If it has been closed, an error occurs and the span is not created.

I don't really have a good idea about what happens to spans that are in-flight during the tracer being closed, however. Couple options -
* Tracer closes immediately (no new spans started), waits some configurable amount of time for spans that have been started and not finished to complete.
* Tracer closes immediately (no new spans started), any unfinished spans are dropped.
* ???

There's probably others I haven't thought of.

rfc/closeable_tracer.md
```diff
@@ -0,0 +1,37 @@
...
+
+Discussions (See [this thread](https://github.com/opentracing/opentracing-java/issues/250) for example) around the inclusion of a `Close` method in Java have crystallized around including this feature in the Java interface. Specifically, there would be value in having `Close` specified on the OpenTracing interface in order to prevent a cast to the concrete tracer, which cannot occur in a situation where the implementation is being loaded dynamically. This situation exists in C# as well, where a similar pattern of try-with-resource (called `using`) exists.
+
+# Specification Changes
+The `Tracer` interface is extended with the following:
+
+* **Close**, a method that is responsible for closing the active tracer. A `Tracer` that has been closed should not be able to record any new spans. For `Tracer`s that record `Span`s for persistance, calling **Close** should flush the current in-memory collection to the configured persistance store. For stateless `Tracer`s or similar, this can be a NoOp. The **Close** method should be considered idempotent; closing an already closed `Tracer` should not emit an error.
```
yurishkuro: s/emit an error/raise an error/

Because I'll most certainly want it to "emit" error, e.g. by logging it, but not by throwing exception.

yurishkuro:
> If it has been closed, an error occurs and the span is not created.

that's not possible - methods for creating a span are not allowed to fail. You can return a noop-equivalent of a span.

> what happens to spans that are in-flight

I think the required guarantee is that spans that have been Finished() prior to tracer.close() call must be flushed. Spans created or finished after tracer.close() may or may not be flushed, but should not be throwing errors and messing up the application.

rfc/closeable_tracer.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Specification Changes
+The `Tracer` interface is extended with the following:
+
+* **Close**, a method that is responsible for closing the active tracer. A `Tracer` that has been closed should not be able to record or create any new spans. In the event that a closed `Tracer` attempts to start a new span, it should return a NoOp-equivalent representation of a `Span`. For `Tracer`s that record `Span`s for persistance, calling **Close** should flush the current in-memory collection to the configured persistance store. For stateless `Tracer`s or similar, this can be a NoOp. The **Close** method should be considered idempotent; closing an already closed `Tracer` should not raise an error. `Span`s that are created or finished after a `Tracer` has been closed may or may not be flushed.
```
yurishkuro: > A `Tracer` that has been closed should not be able to record or create any new spans. In the event that a closed `Tracer` attempts to start a new span, it should return a NoOp-equivalent representation of a `Span`.

I think these two sentences are too prescriptive & unnecessary. The last sentence already says that spans may get lost if created after tracer.close(). I would only say that creating and finishing spans after the tracer was closed should not raise errors to the application, but otherwise I don't see why we need to require any specific behavior. In fact, in the most implementations I would expect a call to close() having no effect on how the spans are created, the only effect is that when span.finish() is called the tracer may decide to ignore that span since its internal reporter/exporter is already terminated.

yurishkuro:
lgtm

rfc/closeable_tracer.md
```diff
@@ -0,0 +1,37 @@
...
+
+Discussions (See [this thread](https://github.com/opentracing/opentracing-java/issues/250) for example) around the inclusion of a `Close` method in Java have crystallized around including this feature in the Java interface. Specifically, there would be value in having `Close` specified on the OpenTracing interface in order to prevent a cast to the concrete tracer, which cannot occur in a situation where the implementation is being loaded dynamically. This situation exists in C# as well, where a similar pattern of try-with-resource (called `using`) exists.
+
+# Specification Changes
+The `Tracer` interface is extended with the following:
+
+* **Close**, a method that is responsible for closing the active tracer. For `Tracer`s that record `Span`s for persistance, calling **Close** should flush the current in-memory collection to the configured persistance store. For stateless `Tracer`s or similar, this can be a NoOp. The **Close** method should be considered idempotent; closing an already closed `Tracer` should not raise an error. `Span`s that are created or finished after a `Tracer` has been closed may or may not be flushed.
```
jpkrohling: Can we add a comment like:

> The `Tracer` interface should extend the platform's idiomatic "close" interface, such as Java's [`Closeable`](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html).

This would allow the tracer to be used with a try-with-resources construct, similar to the C# example you mention afterwards.

austinlparker: I already indicate that Java/C# should use idomatic close in the 'use case' section, not sure if we want to have language-specific parts in the spec description?

jpkrohling: No, I should have read more carefully :-)

austinlparker:
@tedsuo, are you ok with deferring your question about if close should be synchronous so we can get this merged?

tedsuo:
@austinlparker: yes, the RFC basically says that close is synchronous right now; I would just like that to be explicit. I doubt this is a contentious proposal.

austinlparker:
Alright, I called that out specifically. Thanks!

Issue #131, opened 2018-10-10
Remove service tag and replace with a similar concept on Scope.


tylerbenson:
## Background
A while ago, we added `service` as a tag (#119).  This is overly simplistic and has high probability for user confusion.

## Problem
The problem is we scoped it as a tag specifically for a narrow use case, but as implemented, it is trivial for users to use it in ways that can fundamentally break common tracer logic.

## Proposal
I propose we roll back/remove service as a tag, and move the ability to override the service name to the `Scope`.

Perhaps something like this (in java parlance):
```java
Span span = tracer.buildSpan("stuff").start();
try (Scope scope = tracer.activate(span)) {
  try (Scope newScope = scope.withServiceName("foo")) {
    // newScope is now active scope and service name is overridden for any new children spans.
  }
  // new child spans without service name overridden here.
}
```

This also opens the possibility of additional `Scope` scoped tags or other things.

## Consideration
I don't believe any releases have been published with the new `service` tag included, so at a minimum, we should at least revert #119 and https://github.com/opentracing/opentracing-java/pull/287 before they are released to prevent the need for a longer tail deprecation.

adriancole:
Thanks @tylerbenson. I agree the tag convention should be reverted.

I like that this won't accidentally result in a tag "service" which would be hard to grok whether it was an accident of legacy instrumentation or a new thing. It also addresses the first-class nature of service in distributed tracing. Finally, it deals with scoping properly.

One thing I would suggest is keeping this feature on the tracer as opposed to overloading the scope with more methods.

Ex
```java
Span span = tracer.buildSpan("stuff").start();
try (Scope scope = tracer.activateForService(span, "who-I-proxy-for")) {
   // newScope is now active scope and service name is overridden for any new children spans.
}
```

note the span itself here should be only assigned to one service, so this is not only the children but also the activated one itself. There might be some nuance to how this plays out.

Note: the above allows you to switch services multiple times: nested scopes. If that's a feature it should end up in the tests.

An alternative would be to not allow nesting and do something similar to baggage which is write-only.
For example, adding a `serviceName` method to the span builder.

Ex
```java
Span span = tracer.buildSpan("stuff").serviceName( "who-I-proxy-for").start();
try (Scope scope = tracer.activate(span)) {
   // newScope is now active scope and service name is overridden for any new children spans.
}
```

[...]

PR #135, opened 2018-12-06
Revert "Add tag for service"

tylerbenson:
Per discussion on #131.

Reverts opentracing/specification#119

We can add a better replacement in a separate PR, but I wanted to get this removed before it gets too wide spread.


tylerbenson:
I think we've gotten the feedback we needed.  Merging to avoid further confusion.

PR #136, opened 2019-01-18
Add database collection name as a tag.

a-feld:
This PR extends the semantic conventions to include a `db.collection` tag.

@mwear

### DIFF
--- a/semantic_conventions.md
+++ b/semantic_conventions.md
@@ -18,7 +18,7 @@ Span tags apply to **the entire Span**; as such, they apply to the entire timera
 |:--------------|:-----|:-------------------|
 | `component` | string  | The software package, framework, library, or module that generated the associated Span. E.g., `"grpc"`, `"django"`, `"JDBI"`. |
 | `db.instance` | string | Database instance name. E.g., In java, if the jdbc.url=`"jdbc:mysql://127.0.0.1:3306/customers"`, the instance name is `"customers"`. |
-| `db.collection` | string | Database collection name, generally for nosql type databases. E.g., for MongoDB this is the collection. |
+| `db.collection` | string | Database collection name, generally for nosql type databases. E.g., for MongoDB this is the collection. For DynamoDB, this would be TableName. |
 | `db.statement` | string | A database statement for the given database type. E.g., for `db.type="sql"`, `"SELECT * FROM wuser_table"`; for `db.type="redis"`, `"SET mykey 'WuValue'"`. |
 | `db.type` | string | Database type. For any SQL database, `"sql"`. For others, the lower-case database category, e.g. `"cassandra"`, `"hbase"`, or `"redis"`. |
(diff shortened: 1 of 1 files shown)


semantic_conventions.md
```diff
@@ -18,6 +18,7 @@ Span tags apply to **the entire Span**; as such, they apply to the entire timera
 |:--------------|:-----|:-------------------|
 | `component` | string  | The software package, framework, library, or module that generated the associated Span. E.g., `"grpc"`, `"django"`, `"JDBI"`. |
 | `db.instance` | string | Database instance name. E.g., In java, if the jdbc.url=`"jdbc:mysql://127.0.0.1:3306/customers"`, the instance name is `"customers"`. |
+| `db.collection` | string | Database collection name. E.g., for MongoDB this is the collection. For a SQL database, this is the name of the table. |
```
yurishkuro: > For a SQL database, this is the name of the table.

SQL query may join more than one table, how would a single tag work?

a-feld: Great point! Open to any suggestions around this! We can send the collection as a comma delimited set of strings perhaps? Alternatively, we can have this be applicable only to nosql style datastores which do not support joins. Thoughts?

yurishkuro: I guess my point is how much value such tag provides given the SQL join example. Also, instrumentation is often implemented as middleware that may only get access to the query string, and wouldn't know how to extract the table name. But in nosql drivers it may still be useful, eg in elastic api the caller specifies index names explicitly.

semantic_conventions.md
```diff
@@ -18,6 +18,7 @@ Span tags apply to **the entire Span**; as such, they apply to the entire timera
 |:--------------|:-----|:-------------------|
 | `component` | string  | The software package, framework, library, or module that generated the associated Span. E.g., `"grpc"`, `"django"`, `"JDBI"`. |
 | `db.instance` | string | Database instance name. E.g., In java, if the jdbc.url=`"jdbc:mysql://127.0.0.1:3306/customers"`, the instance name is `"customers"`. |
+| `db.collection` | string | Database collection name, generally for nosql type databases. E.g., for MongoDB this is the collection. |
```
pavolloffay: Could you please provide more examples for different databases so we know we are not introducing a tag only for mongo?

pavolloffay: Enough to put it here as a comment

yurishkuro:
So other examples:
- mongo : collection
- es : index
- Cassandra : table
- no-join SQL query : table
- hive : table
- Pinot : table

Do we see the pattern?

devinsba:
I'll add another one for Yuri's case:

- AWS DynamoDB: table

The exceptions I've found in public cloud are:

- Google FireStore: collection
- Google Cloud Datastore: kind
- Azure Cosmos: collection, graph, or table (depending on mode)

PR #156, opened 2020-08-25
Fix typo

quaff:
"modelling" should be "modeling", see https://en.wikipedia.org/wiki/Data_modeling


yurishkuro:
Both are acceptable.Double-L is the preferred spelling in all English-speaking countries apart from the United States.

PR #157, opened 2020-09-16
Add "slow" standard tag

quaff:
see https://github.com/opentracing-contrib/java-jdbc/pull/99#issue-487826199


yurishkuro:
You do not need a new semantic convention to add a custom tag. This tag is redundant, you can derive the same from already existing data on the span, e.g. duration.

